PR #1, opened 2024-05-28
Bump Version to 0.3.22

emin-grbo:


.github/workflows/changelog-check.yml
```diff
@@ -5,9 +5,6 @@ on:
   pull_request:
     paths:
       - 'version.txt'
-  push:
-    paths:
-      - 'version.txt'
```
e-kononenko, line 10: why do we remove it?

emin-grbo: We have no need to run this on push anymore. Only PR 🙌
I discussed this with @michpohl recently.

e-kononenko (approved):

PR #8, opened 2024-05-29
Remove scopes that aren't needed from documentation

michpohl:
For client credentials authentication, the user doesn't need scopes, so it's misleading to reference them in the documentation. Therefore this removes them.


asendra:
Closing as it has been addressed in a parallel PR 👍

PR #9, opened 2024-05-29
Moved PlayLog tests to a separate GH Action

asendra:
While we continue development of the PlayLog unit tests, we have moved them to a separate action that we can trigger manually.

Once they are completed and stable, we will decide the frequency on which to run them, given that they are inherently slow.

### DIFF
--- a/.github/workflows/playlog-tests.yml
+++ b/.github/workflows/playlog-tests.yml
@@ -2,6 +2,8 @@ name: Run PlayLog Unit Tests
 
 on:
   workflow_dispatch:
+  schedule:
+    - cron: "5 3 * * *"
 
 jobs:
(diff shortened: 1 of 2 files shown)


ziad-halabi9 (approved):

emin-grbo (approved):

ceciliasaraiva (approved):
Nice! What do you think about having a nightly run until we have they all fixed? It can help with finding the worst ones also when going back to work on them.

asendra:
> Nice! What do you think about having a nightly run until we have they all fixed? It can help with finding the worst ones also when going back to work on them.

Sounds like a great idea! We will be able to see if tests are also flaky when running alone.

PR #10, opened 2024-05-29
Fix test flakiness in PlayerEventSender

asendra:
This PR introduces the concept of an `AsyncScheduler` protocol to control async tasks inside `PlayerEventSender` in tests.

It also modifies the tests in `EventsTests` and `PlayerEventSenderTests` to remove `optimizedWait` and use this new scheduler, when possible. (tests that use the `shouldSendEventsInDeinit` flag can't be controlled like this because what is flaky is the autorelease pool)

`PlayereventSenderTests` also had tests duplicated to test both using `deinit` and not using `deinit`, which is not needed because that code path is not touched by those tests.


emin-grbo (approved):

emin-grbo:
Stellar work Alberto, this is amazing!

Sources/Player/Common/World/AsyncSchedulerFactoryProvider.swift
```diff
@@ -0,0 +1,13 @@
+import Foundation
+
+// MARK: - AsyncSchedulerFactoryProvider.swift
+
+struct AsyncSchedulerFactoryProvider {
+	var newFactory: () -> AsyncSchedulerFactory
+}
+
+extension AsyncSchedulerFactoryProvider {
```
ceciliasaraiva, line 9: [suggestion] not a big deal, but maybe we can follow others we have and split into 2 files, and keep the live one in `AsyncSchedulerFactoryProvider+Live`.

Sources/Player/PlaybackEngine/Internal/Events/AsyncScheduler.swift
```diff
@@ -0,0 +1,23 @@
+protocol AsyncScheduler {
```
ceciliasaraiva, line 1: [suggestion] What do you think about splitting each one in their own file for better discoverability

Sources/Player/PlaybackEngine/Internal/Events/AsyncScheduler.swift
```diff
@@ -0,0 +1,23 @@
...
+}
+
+protocol AsyncSchedulerFactory {
+  @discardableResult
+  func create(code: @escaping () async -> Void) -> AsyncScheduler
+}
+
+extension Task: AsyncScheduler {
+  static func schedule(code: @escaping () async -> Void) -> AsyncScheduler {
+	Task<Void, Never> {
```
ceciliasaraiva, line 12: [suggestion] I think we can use our typealias `SafeTask`. Same below.
[suggestion 2] I think formatting is missing, I think you just need to run the setup again.
```suggestion
	SafeTask {
```

Sources/Player/PlaybackEngine/Internal/Events/AsyncScheduler.swift
```diff
@@ -0,0 +1,23 @@
...
+	  await code()
+	}
+  }
+}
+
+class TaskAsyncSchedulerFactory: AsyncSchedulerFactory {
```
ceciliasaraiva, line 18: [suggestion] it can be `final`, no?
```suggestion
final class TaskAsyncSchedulerFactory: AsyncSchedulerFactory {
```

Sources/Player/PlaybackEngine/Internal/Events/PlayerEventSender.swift
```diff
@@ -77,10 +80,9 @@ class PlayerEventSender {
 	}
 
 	func send(_ offlinePlay: OfflinePlay) {
-		SafeTask { [weak self] in
-			guard let self else {
-				return
-			}
+		asyncSchedulerFactory.create { @MainActor [weak self] in
```
ceciliasaraiva, line 83: [question] do we need the `@MainActor` here (and others below)?

Tests/PlayerTests/EventsTests.swift
```diff
@@ -154,12 +166,14 @@ final class EventsTests: XCTestCase {
...
 	}
 
-	func testCsssCpbiAndCsseSentAfterSuccessfulLoadAndReset() {
+	func testCsssCpbiAndCsseSentAfterSuccessfulLoadAndReset() async {
 		shouldUseEventProducer = true
 		PlayerWorld.developmentFeatureFlagProvider.shouldSendEventsInDeinit = false
 
 		JsonEncodedResponseURLProtocol.succeed(with: Constants.trackPlaybackInfo)
 
+		await self.asyncSchedulerFactory.executeAll()
```
ceciliasaraiva, line 175: [suggestion] I think in this whole file we don't need the async scheduler since we use a mock and not the actual one (different from `PlayerEventSenderTests`). 🤔 If there were flaky, maybe it is something else we need to look at.

Tests/PlayerTests/Mocks/PlaybackEngine/Internal/Events/AsyncschedulerSpy.swift
```diff
@@ -0,0 +1,43 @@
+import Foundation
+@testable import Player
+
+class AsyncSchedulerSpy: AsyncScheduler {
```
ceciliasaraiva, line 4: [suggestion] it can be `final`, right?
```suggestion
final class AsyncSchedulerSpy: AsyncScheduler {
```
[suggestion 2] Small camel case typo in the class name: `AsyncschedulerSpy`

Tests/PlayerTests/Mocks/PlaybackEngine/Internal/Events/AsyncschedulerSpy.swift
```diff
@@ -0,0 +1,43 @@
...
+	static func schedule(code: @escaping () async -> Void) -> AsyncScheduler {
+		return AsyncSchedulerSpy(code: code)
+	}
+	
+	func execute() async {
+		await code?()
+	}
+}
+
+class AsyncSchedulerFactorySpy: AsyncSchedulerFactory {
```
ceciliasaraiva, line 20: [suggestion] it can be `final`, right?
```suggestion
final class AsyncSchedulerFactorySpy: AsyncSchedulerFactory {
```
[suggestion 2] we could also split them in different files for better discoverability.

Tests/PlayerTests/Player/PlaybackEngine/Internal/Events/PlayerEventSenderTests.swift
```diff
@@ -116,41 +124,7 @@ extension PlayerEventSenderTests {
...
-			name: StreamingMetricNames.streamingSessionStart,
-			group: EventGroup.streamingMetrics,
-			payload: streamingSessionStart
-		)
-	}
-
-	func test_send_StreamingMetricsEvent_with_StreamingSessionStart_legacy2() {
-		PlayerWorld.developmentFeatureFlagProvider.shouldSendEventsInDeinit = true
-
+	func test_send_StreamingMetricsEvent_with_StreamingSessionStart_legacy() async {
```
ceciliasaraiva, line 127: [suggestion] I will only write here but it is valid for each set of test which we assert the same. Basically i had added 4 for each set because we have to cover the 2 existing conditions which result in 4 different combinations:
- send events in deinit and use EP
- send events in deinit and don't use EP
- don't send events in deinit and use EP
- don't send events in deinit and don't use EP
Let me know if I totally missed something. 🙈

ceciliasaraiva (commented):
Great work with the async scheduler so we can have more reliable tests for PlayerEventSender! 🙌 Thanks for that!
I've added some comments, I do apologize beforehand for the high number (some are minor). 😅

Tests/PlayerTests/Player/PlaybackEngine/Internal/Events/PlayerEventSenderTests.swift
asendra: That is correct for `EventsTests` as they are testing at the `PlayerEngine` level.

In `PlayerEventSenderTests` we are jsut creating mock objects and sending them explicitly, so it doesn't test anything about `deinit` vs no `deinit` (It's `PlayerItem.deinit`, and there are no PlayerItems here)

Sources/Player/Common/World/AsyncSchedulerFactoryProvider.swift
asendra: I actually just saw that we don't separate any of the Live providers we use in `PlayerWorld` but we have to start somewhere :)

asendra: That's a lie. We do on `DevelopmentFeatureFlagProvider` haha

Sources/Player/PlaybackEngine/Internal/Events/AsyncScheduler.swift
asendra: Done!

Sources/Player/PlaybackEngine/Internal/Events/PlayerEventSender.swift
asendra: This was a good catch @ceciliasaraiva! 👏

Tests/PlayerTests/EventsTests.swift
asendra: Oh, you are right!
I don't know if these specifically were flaky. The `PlayerEventSenderTests` I'm sure. I did it here just in case.
It seems like in `EventsTests` we only need the `optimizedWait()` for the cases where `deinit` is used.

Sources/Player/PlaybackEngine/Internal/Events/AsyncScheduler/AsyncScheduler+Task.swift
```diff
@@ -0,0 +1,7 @@
+extension Task: AsyncScheduler {
```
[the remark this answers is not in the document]
asendra: lol 😆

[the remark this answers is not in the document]
asendra: yeah, I had not done `./local-setup.sh` since the repo migration.

PR #11, opened 2024-05-29
Fix test flakiness in StreamingPrivilegesHandlerTests

asendra:
This PR improves speed and stability for StreamingPrivilegesHandlerTests by implementing a more efficient and flexible wait mechanism to allow the async tasks to do their work.

This is not fully controllable, but it has been proven to be quite reliable for us in other tests, and much better than waiting a set amount of time.

### Before:

Screenshot 2024-05-29 at 16 29 00

### After:
Screenshot 2024-05-29 at 16 29 26

### 100 runs:
Screenshot 2024-05-29 at 16 10 08

### DIFF
--- a/Tests/PlayerTests/Helpers/XCTestCase+Player.swift
+++ b/Tests/PlayerTests/Helpers/XCTestCase+Player.swift
@@ -1,5 +1,6 @@
 import Foundation
 import XCTest
+@testable import Player
 
 extension XCTestCase {
 	func optimizedWait(timeout: Double = 10.0, step: Double = 0.1, until condition: () -> Bool) {
@@ -7,6 +8,21 @@ extension XCTestCase {
 		while timer < timeout {
 			let expectation = expectation(description: "Optimized wait: Waiting for condition to be met")
 			_ = XCTWaiter.wait(for: [expectation], timeout: step)
+
+			if condition() {
+				return
+			}
+
+			timer += step
+		}
+
+		XCTFail("Optimized wait: Condition was not met")
+	}
+	
+	func asyncOptimizedWait(timeout: Double = 10.0, step: Double = 0.1, until condition: @escaping () -> Bool) async {
+		var timer: Double = 0
+		while timer < timeout {
+			try? await Task.sleep(seconds: step)
 
 			if condition() {
 				return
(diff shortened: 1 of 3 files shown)


emin-grbo (approved):
Impressive! Execution speed is becoming more and more important and this is a perfect boost it. Great work!

Tests/PlayerTests/Helpers/XCTestCase+Player.swift
```diff
@@ -17,4 +18,19 @@ extension XCTestCase {
 
 		XCTFail("Optimized wait: Condition was not met")
 	}
+	
+	func asyncOptimizedWait(timeout: Double = 10.0, step: Double = 0.1, until condition: @escaping () -> Bool) async {
+		var timer: Double = 0
+		while timer < timeout {
+			try? await Task.sleep(seconds: step)
```
ceciliasaraiva, line 25: [question] is the sleep less flaky than the `XCTWaiter.wait`?
[question2] Have you revisited the default timeout we have been using? (To know if timeout can be shorter, I remember we had increased but maybe after improvements it is not needed so long anymore)

ceciliasaraiva (approved):
Nice, we haven't touched these for a while so they still had the long sleep's 💤 Just added 2 questions.

ceciliasaraiva:
Maybe it's worth executing in the CI many times to make sure they are not flaky anymore. 🤓

Tests/PlayerTests/Helpers/XCTestCase+Player.swift
asendra: I really can't say what they do under the hood. I would expect both to be the same with respect to flakiness.
But `Task.sleep` feels like a better Primitive for what we are doing. `XCTWaiter.wait` has extra functionality, as it also allows for the expectation to be fullfilled earlier, and you can check the result to know if it timed out or was fullfilled. And we don't really need any of that here.

Compared to `XCTWaiter.wait` outside `optimizedWait` with a fixed time, this is more robust because we can use bigger timeouts without generally affecting the test speed.

I don't think the default timeout matters much, because we check continuously to exit early.

asendra:
> Maybe it's worth executing in the CI many times to make sure they are not flaky anymore. 🤓

mm yeah, although we don't have a way to run the tests manually on CI. I can add that that.

PR #17, opened 2024-06-10
Part 1: Feature flag setup for new caching implementation

asendra:
This PR adds a new FeatureFlag to control access to the new player caching implementations that will follow in the next PR.

Renames the 3-4 classes for the current AVQueuePlayerWrapper implementation as `ClassNameLegacy` as we have done in previous similar situations. This duplicates code but will allow us to quickly delete the legacy implementation once we have validated the new one in production.

It also adds metrics tracking via a new `SessionTag`. 

All these will allows us to control the deployment of the new code and measure its impact vs the previous implementation.

### DIFF
--- a/Sources/Player/Common/Extensions/AVPlayer.swift
+++ b/Sources/Player/Common/Extensions/AVPlayer.swift
@@ -1,0 +1,16 @@
+import Foundation
+import MediaPlayer
+
+extension AVPlayer {
+	func seek(to position: Double) async -> Bool {
+		await withCheckedContinuation { continuation in
+			guard let currentItem else {
+				return
+			}
+
+			seek(to: CMTime(seconds: min(position, currentItem.duration.seconds), preferredTimescale: 1000)) { finished in
+				continuation.resume(returning: finished)
+			}
+		}
+	}
+}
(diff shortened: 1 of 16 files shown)


Sources/Player/Common/Extensions/AVPlayer.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+import MediaPlayer
+
+extension AVPlayer {
+	func seek(to position: Double) async -> Bool {
```
ceciliasaraiva, line 5: [question] why have we moved this one out of the its file?

Sources/Player/Common/Extensions/AVURLAsset.swift
```diff
@@ -0,0 +1,8 @@
+import Foundation
+import MediaPlayer
+
+extension AVURLAsset {
```
ceciliasaraiva, line 4: [question] and this one?

ceciliasaraiva (approved):
nice, just two minor questions

Sources/Player/Common/Extensions/AVPlayer.swift
asendra: Same answer to both!
Both extensions were part of files that I'm duplicating, so I moved them out to their own file so I could use them from both implementations without errors from duplicated symbols.

Sources/Player/Common/Extensions/AVURLAsset.swift
asendra: Same

PR #18, opened 2024-06-10
Part 2: New Player caching implementation

asendra:
This Pr adds the new caching implementation that we have already validated in internal testing builds, behind the Feature Flag set up in the previous PR, for extra control.

The only small change outside the FF is the changes to the `Asset` interface, to make it compatible with both implementations. this change is minimal though.


Sources/Player/Common/Extensions/String+Random.swift
asendra, line 1: @ceciliasaraiva in the old PR you mentioned using a substring of UUID instead but I believe UUID does not guarantee the same uniqueness for substrings, and the full UUID would be too long? 
I can change it to a UUID if you think it is important.

Sources/Player/Common/UserDefaults/UserDefaultsClient+Live.swift
asendra, line 1: `public` because our externalPlayer needs it.

Sources/Player/Common/UserDefaults/UserDefaultsClient.swift
asendra, line 1: `public` because our externalPlayer needs it.

Sources/Player/PlaybackEngine/Internal/Asset.swift
```diff
@@ -8,9 +8,9 @@ open class Asset {
 	let player: GenericMediaPlayer
 	private var loudnessNormalizationConfiguration: LoudnessNormalizationConfiguration
 
-	@Atomic var assetPosition: Double
+	@Atomic var assetPosition: Double = 0
 	@Atomic var playbackMetadata: AssetPlaybackMetadata?
-	@Atomic private var cached: Bool = false
+	@Atomic var cacheState: AssetCacheState?
```
asendra, line 13: This is the only change not behind the FF but the change in the Legacy implementation to this new interface is very minor.

Sources/Player/PlaybackEngine/PlayerEngine.swift
```diff
@@ -228,6 +228,10 @@ final class PlayerEngine {
 				return
 			}
 
+			guard self.nextItem?.mediaProduct != mediaProduct else {
+				return
+			}
```
asendra, line 233: This helps filter out the extra `setNext:` calls that the client is doing.
I've done the change in here because it helps improve the caching logic (repeated setNext calls to the same were causing some issues)

Sources/Player/Common/Extensions/String+Random.swift
ceciliasaraiva: I have the feeling it's more unique than this setup, but that's just my thought (I guess the only difference in terms of characters is the hifen in UUID). The guarantee of UUID is for the full length, so either option we need to accept the possibility of collisions. I would go for UUID because I would guess its internal logic has a more robust way of generating the unique identifier than the random func - which is pseudo random anyway.

asendra: done!

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AssetCache.swift
```diff
@@ -0,0 +1,65 @@
...
+
+public final class AssetCache {
+	private var userDefaults: UserDefaultsClient
+
+	public init(userDefaults: UserDefaultsClient = UserDefaultsClient.live()) {
+		self.userDefaults = userDefaults
+	}
+
+	public func get(_ key: String) -> AssetCacheState {
+		var bookmarkIsStale = false
```
ceciliasaraiva, line 30: [suggestion] I think this alternative reads a bit more as an assertion:
```suggestion
		var isBookmarkStale = false
```

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AssetCache.swift
```diff
@@ -0,0 +1,65 @@
...
+		let state = get(key)
+		userDefaults.removeObjectForKey(key)
+		if case let .cached(url) = state.status {
+			deletefile(at: url)
+		}
+	}
+}
+
+extension AssetCache {
+	func deletefile(at url: URL) {
```
ceciliasaraiva, line 61: [suggestion] small typo:
```suggestion
	func deleteFile(at url: URL) {
```

Sources/Player/PlaybackEngine/Internal/Asset.swift
```diff
@@ -50,16 +49,16 @@ open class Asset {
 		self.playbackMetadata = playbackMetadata
 	}
 
-	/// Method to obtain the current caching status
-	/// - Returns: The caching status
-	public func isCached() -> Bool {
-		cached
+	/// Method to get the current caching state
+	/// - Returns: The asset cache state
+	public func getCacheState() -> AssetCacheState? {
```
ceciliasaraiva, line 54: [Question] when could this be optional and when would we actually like for it to be optional instead of not cached?

Sources/Player/PlaybackEngine/PlayerEngine.swift
ceciliasaraiva: How about adding this one behind a dev FF or actual FF? (Not sure there could be some scenario we are missing to make it risky. Maybe the only one that needs testing is when we repeat 🤔)

Sources/Player/PlaybackEngine/PlayerEngine.swift
```diff
@@ -253,6 +257,26 @@ final class PlayerEngine {
...
 
+	func resetOrUnload() {
+		if featureFlagProvider.shouldUseImprovedCaching() {
+			unload()
+		} else {
+			reset()
+		}
+	}
+
+	func unload() {
```
ceciliasaraiva, line 268: [suggestion] I believe we have briefly discussed this already, regarding maybe adding some explicit comments on the difference between the unload and reset, which from first look it is not super clear.

Tests/PlayerTests/Mocks/PlaybackEngine/Internal/PlayerLoaderMock.swift
```diff
@@ -9,6 +9,7 @@ final class PlayerLoaderMock: PlayerLoader {
 	private(set) var loadPlayableStorageItems = [PlayableStorageItem]()
 	private(set) var loadStoredMediaProducts = [StoredMediaProduct]()
 	private(set) var loadPlaybackInfos = [PlaybackInfo]()
+	private(set) var unloadCallCount = 0
```
ceciliasaraiva, line 12: [question] Could we have a simple test of PlayerEngine, which mocks playing and then after we call unload, it will check that this count is 1 and reset remains 0? I think it would be basically copy and paste of similar tests we already have, it doesn't really get even close to testing all changes here, but a tiny tiny bit at least, and it doesn't require further changes. 😅

ceciliasaraiva (approved):
That review took a while 🙌 Nice work, I've added some comments, nothing major I believe.

Sources/Player/PlaybackEngine/PlayerEngine.swift
asendra: I can put it behind the same FF as the new caching behavior so at least they are tied and we can revert to the current state. 
We have lots of flags active that we need to start confirming that everything is ok and removing 😅

Sources/Player/PlaybackEngine/Internal/Asset.swift
asendra: Optional will be when we don't want to cache it. (It is a stored file or some format/quality we don't want cached)
That logic is done at a higher level (`InternalPlayerLoader`) to avoid propagating extra logic to the player that is not needed.

PR #20, opened 2024-06-13
Fix long buffering w/ new caching implementation

asendra:
## Context

While testing the new caching implementation currently under a FeatureFlag, it was noticed that in some scenarios, there was a 3-4 second buffering that should not be happening.

## Description

When the first track in a PlayerEngine instance is loaded, and the track is not offlined or cached, it could lead to the track not starting playback until the full track is downloaded/cached.

The solution is to load into the player queue on the play() command whatever track has been loaded but was being cached.


emin-grbo (approved):

ceciliasaraiva (approved):
nice catch!

PR #21, opened 2024-06-13
Fix playback not stopping immediately on reset

asendra:
## Context

While testing the new caching implementation currently under a FeatureFlag, it was noticed that when killing the app, playback would continue for 3-4 extra seconds instead of stopping automatically.

## Description

A change was made to call `player.removeAllItems()` instead of `player.pause()` when unloading/resetting the player.
This is what the documentation for removeAllItems() says
```
/**
  @method     removeAllItems
  @abstract   Removes all items from the queue.
  @discussion Stops playback by the target.
*/
open func removeAllItems()
```

The solution is to call `player.pause()` too as `player.removeAllItems()` is not enough to stop playback, even if the documentation says it should be.


ceciliasaraiva (approved):
simple and effective 🤓

PR #24, opened 2024-06-13
Ensure queue is accessed on background

emin-grbo:
If user has selected to block non-essential cookies, there were instances where writing to the DB caused collisions, which resulted in events being dropped.

This PR introduces a `Configuration` with `.qos` set to `background` which mitigates this issue.

Major props to @micai23 for helping it debugging this and finding the actual solution 🙌 🎉

### DIFF
--- a/Sources/EventProducer/Events/Logic/EventQueue.swift
+++ b/Sources/EventProducer/Events/Logic/EventQueue.swift
@@ -4,7 +4,6 @@ import SWXMLHash
 
 final class EventQueue {
 	static let databaseName = "EventQueueDatabase.sqlite"
-
 	static let shared = EventQueue()
 	var dbQueue: DatabaseQueue!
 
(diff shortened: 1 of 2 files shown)


asendra (commented):
Are you sure it is related with the non-essential cookies thing that you mention? I don't understand how that could be.
Also, QoS has nothing to do with DB locking... and mitigation is not good enough, we can't drop events.

I believe the error is this silly mistake: https://github.com/tidal-music/tidal-sdk-ios/pull/26

What do you think?

asendra:
Also, just as a small nitpick with the title. 

`configuration.qos = .background ` has nothing to do with the queue being accessed in the background or any thread in specific.

It only means that the thread accessing the queue has less priority.

It is the same naming that Apple uses in lots of APIs [like this one](https://developer.apple.com/documentation/swift/taskpriority)

emin-grbo:
Addressed properly in: https://github.com/tidal-music/tidal-sdk-ios/pull/26#pullrequestreview-2118247448

PR #25, opened 2024-06-13
Ensure cache delegate updates happen in the Player queue

asendra:
Noticed a likely issue after our new caching implementation where the cache delegate updates do not use the player queue to update the internal state, which could cause some issues or even crashes.


ceciliasaraiva (approved):
🚀

PR #26, opened 2024-06-13
Fixed EventProducer SQLITE_LOCKED errors

asendra:
I believe a typo was causing the `EventProducer` module to open multiple connections to the same DB, causing `SQLITE_LOCKED` errors.

### DIFF
--- a/Sources/EventProducer/Monitoring/MonitoringQueue.swift
+++ b/Sources/EventProducer/Monitoring/MonitoringQueue.swift
@@ -11,7 +11,7 @@ public final class MonitoringQueue {
 	}
 
 	private func setup() throws {
-		guard let databaseURL = FileManagerHelper.shared.eventQueueDatabaseURL else {
+		guard let databaseURL = FileManagerHelper.shared.monitoringQueueDatabaseURL else {
 			throw EventProducerError.monitoringQueueDatabaseURLFailure
 		}
 


micai23 (approved):
Wow hawk eyes 👀 good catch 🙌

Sources/EventProducer/Monitoring/MonitoringQueue.swift
```diff
@@ -11,7 +11,7 @@ public final class MonitoringQueue {
 	}
 
 	private func setup() throws {
-		guard let databaseURL = FileManagerHelper.shared.eventQueueDatabaseURL else {
+		guard let databaseURL = FileManagerHelper.shared.monitoringQueueDatabaseURL else {
```
micai23, line 14: @emin-grbo I would test this also and see. 
1) try reverting the changes, use @asendra fix and see what happens 
2) try with both changes too

asendra:
> Wow hawk eyes 👀 good catch 🙌

My eyes are not that good 😆 

DB locking sounded like a concurrency issue, so I just read their [Concurrency guide](https://swiftpackageindex.com/groue/grdb.swift/v6.27.0/documentation/grdb/concurrency) again.

The first point mentions making sure there is only one connection per DB, ss I searched the project for the database path and bingo 😅

micai23:
> > Wow hawk eyes 👀 good catch 🙌
> 
> My eyes are not that good 😆
> 
> DB locking sounded like a concurrency issue, so I just read their [Concurrency guide](https://swiftpackageindex.com/groue/grdb.swift/v6.27.0/documentation/grdb/concurrency) again.
> 
> The first point mentions making sure there is only one connection per DB, so I searched the project for the database path and bingo 😅

When we were debugging, we were only able to reproduce the error without breakpoints, it never happened with breakpoints that's why I thought it was a threading issue.

asendra:
> > > Wow hawk eyes 👀 good catch 🙌
> > 
> > 
> > My eyes are not that good 😆
> > DB locking sounded like a concurrency issue, so I just read their [Concurrency guide](https://swiftpackageindex.com/groue/grdb.swift/v6.27.0/documentation/grdb/concurrency) again.
> > The first point mentions making sure there is only one connection per DB, so I searched the project for the database path and bingo 😅
> 
> When we were debugging, we were only able to reproduce the error without breakpoints, it never happened with breakpoints that's why I thought it was a threading issue.

It is a threading issue. (Two simultaneous connections to the database)

The QoS change (which stands for Quality of Service) only affects the priority of the internal queue used for the DB connection.
It does not affect the threading architecture in any way.

Most likely the reason why you saw improvements, is because you lowered the QoS for one of the two connections, effectively throttling one of them, which makes sense that it would result in fewer "collisions" with the other connection and fewer errors.

[Here you can see](https://github.com/groue/GRDB.swift/blob/4b934fda754b1cab34394ea89c2dd5acd170d50e/GRDB/Core/Configuration.swift#L475) that the `.qos` parameter is only used as a config parameter for the DispatchQueues used to read and write to the DB.

[And here you can see](https://developer.apple.com/documentation/dispatch/dispatchqueue/2300077-global) the documentation for DispatchQueue.

emin-grbo:
Oh wow 🤦‍♂️ We were actually changing the name of the `static let databaseName =` as it did sound like this issue, and was very surprised it did not work. Only to see now it was never actually used!
Stellar work spotting this.

I will of course re-test but trust this is it as it really explains the reported issue very well 🙌

emin-grbo (approved):
Yup, that did it!
Well done!

PR #29, opened 2024-06-18
Add new empty catalog module

michpohl:
Only contains the generated "Hello, World" code so far.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -52,6 +52,18 @@ let package = Package(
 			]
 		),
 		.target(
+			name: "Catalog",
+			plugins: [
+				.plugin(name: "SwiftLint", package: "SwiftLintPlugin"),
+			]
+		),
+		.testTarget(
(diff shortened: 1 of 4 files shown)


asendra (approved):
A bit more extra changes will be needed to add Documentation support for this package.

Currently the documentation setup is a bit of wonky, which is why I did not add it to the `generate-module.sh` script.

But I think we can live with it until the module is further along, right?

ceciliasaraiva (approved):

michpohl:
> A bit more extra changes will be needed to add Documentation support for this package.
> 
> Currently the documentation setup is a bit of wonky, which is why I did not add it to the `generate-module.sh` script.
> 
> But I think we can live with it until the module is further along, right?

I had to do a bunch of things manually here - will look into that at a later time. Currently it is just about providing a few things so folks aren't blocked.

PR #30, opened 2024-06-19
Update README.md

micai23:
Readme had `pl` Polish language instead of english for  creating modules using `PascalCase`
Fix by removing language code and letting browser determine display language.


emin-grbo (approved):
Good spot!

PR #34, opened 2024-07-01
Add generated catalogue models

michpohl:
This PR adds modules generated from the `catalogue` endpoint's `openapi` definition.

Changes included:
- Add api definition and generator config in `./SourcesCatalogue/openapi`
- Add generated models in `./Sources/Generated/Models`
- Exclude `Generated` folder from swiftlint checks

This is only part of the code generated by `openapi-generator`. But as there are still open questions as to what the final catalogue module should look like, only the models will be added at this point - the rest to follow once we have the full design spec.


asendra:
Unfortunately @bkdl is out, but he already tried to use the openapi-generator for the Creator content project and ran into some issues.

I think it was something about a module depending on another module that auto generated the code not having access to the code (had not been generated yet)

In this case, as you are committing the generated code, I think it should be fine. Was that the reason you choose to commit it? Is not the default recommendation in their FAQ
https://swiftpackageindex.com/apple/swift-openapi-generator/1.2.1/documentation/swift-openapi-generator/frequently-asked-questions#Do-I-commit-the-generated-code-to-my-source-repository

michpohl:
> Unfortunately @bkdl is out, but he already tried to use the openapi-generator for the Creator content project and ran into some issues.
> 
> I think it was something about a module depending on another module that auto generated the code not having access to the code (had not been generated yet)
> 
> In this case, as you are committing the generated code, I think it should be fine. Was that the reason you choose to commit it? Is not the default recommendation in their FAQ https://swiftpackageindex.com/apple/swift-openapi-generator/1.2.1/documentation/swift-openapi-generator/frequently-asked-questions#Do-I-commit-the-generated-code-to-my-source-repository

This is a super early-stage commit. We agreed that for now, we would just exposer the model classes so they can be used if needed, for preparing the integration.

We need to first figure out what exactly the final result is supposed to look like, and hopefully the code generation will play along. So in the future they might be removed again in favor of an automated step.

mercedesbunz:
> Unfortunately @bkdl is out, but he already tried to use the openapi-generator for the Creator content project and ran into some issues.
> 
> I think it was something about a module depending on another module that auto generated the code not having access to the code (had not been generated yet)
> 
> In this case, as you are committing the generated code, I think it should be fine. Was that the reason you choose to commit it? Is not the default recommendation in their FAQ [swiftpackageindex.com/apple/swift-openapi-generator/1.2.1/documentation/swift-openapi-generator/frequently-asked-questions](https://swiftpackageindex.com/apple/swift-openapi-generator/1.2.1/documentation/swift-openapi-generator/frequently-asked-questions#Do-I-commit-the-generated-code-to-my-source-repository)

This PR is using a different open api generator. It doesn't provide a plugin and besides I think not polluting the build step with building the files every time makes sense. 
the generated files from this generator also are nicer split apart

mercedesbunz (approved):

ceciliasaraiva (commented):
Hey Michael, do you have swiftformat set up? Since I see different indentation being used here. 🤔

emin-grbo (approved):

PR #35, opened 2024-07-04
Fix event producer crashes

e-kononenko:
Sometimes accessing to `DatabaseQueue` fails, looks like it is not initialized properly and we have a crash because of this. One reason can be that "protected data" is not yet available when we want to send an event. 

The problems in the current implementation that prevented us from solving the issue (or at least to understand the reason) are:

- Singletons are widely used across different classes, we have do not really control entities initialization
- `DatabaseQueue` is initialized during main class (`TidalEventSender`) initialization. It will try to access the database, but if something goes wrong - we do not throw any error and just ignore this situation. 
- `DatabaseQueue` variables are implicitly unwrapped optionals, if their initialization fails, we just can have a crash when trying to access them (in `EventQueue` and `MonitoringQueue` classes)
- Also, the way our schedulers work is a mix and match between different async techniques. This can also be a reason of this or other crashes. 

So, in this PR I am trying to solve all these issues.

In particular, I did:
- Got rid of the singletons in `EventQueue`, `EventSubmitter`, `Monitoring`, `MonitoringQueue`. Got rid of usages of the singletons in related functions and variables.
- Initialized all these entities as instance variables in `TidalEventSender` and pass them as arguments where it is needed.
- Added convenience initializers in schedulers (`EventScheduler`, `MonitoringScheduler`), so that they can be initialized with config, where when can take other parameters from. This can help us to extend config if needed without rewriting existing code in other classes.
- `EventQueue` and `MonitoringQueue`:
  -  Initialized underlying `DatabaseQueue` lazily with throwing an error if something goes wrong. Thus, the database will be initialized only when we need it (when we attempt to access it first time), so we have more time until protected data is available. 
  - Got rid of implicit unwrapping there (`var dbQueue: DatabaseQueue!`). If the database is already initialized - we just return it, otherwise we try to initialize it. 
  - `DatabaseQueue` initialization can throw an error and we don't swallow it anymore. Instead, this error will be thrown by a method which tries to access it, so clients can properly handle it.
- New `Scheduler` protocol, which extension implements scheduling logic purely in async/await. `EventScheduler` and `MonitoringScheduler` now conform to the protocol instead of having a repeated scheduling logic.
- Adjusted unit tests to reflect the changes mentioned above.

### DIFF
--- a/Sources/EventProducer/Common/Scheduler.swift
+++ b/Sources/EventProducer/Common/Scheduler.swift
@@ -7,6 +7,10 @@ protocol Scheduler: AnyObject {
 }
 
 extension Scheduler {
+	/// The Scheduler applies a best-effort approach to sending events to the TL Consumer, meaning that it continuously tries to send
+	/// events while the app is executing,
+	/// regardless of current internet conditions or if the app considers itself to be online or not. As long as the app is
+	/// executing, the Scheduler shall be running
 	func runScheduling(with headerHelper: HeaderHelper) {
 		guard schedulerTask == nil else {
 			return
--- a/Sources/EventProducer/Common/Scheduler.swift
+++ b/Sources/EventProducer/Common/Scheduler.swift
@@ -1,0 +1,33 @@
+import Foundation
+
+protocol Scheduler: AnyObject {
+	var schedulingTime: TimeInterval { get }
+	func timerTriggered(headerHelper: HeaderHelper) async throws
+	var schedulerTask: Task<Void, Error>? { get set }
+}
+
+extension Scheduler {
+	func runScheduling(with headerHelper: HeaderHelper) {
+		guard schedulerTask == nil else {
+			return
+		}
+		
+		schedulerTask = Task { [weak self, headerHelper] in
+			repeat {
+				guard let self else {
+					return
+				}
+				
+				let schedulingTimeInNanoseconds = UInt64(schedulingTime) * NSEC_PER_SEC
+				
+				try await Task.sleep(nanoseconds: schedulingTimeInNanoseconds)
+				try await self.timerTriggered(headerHelper: headerHelper)
+			} while !Task.isCancelled
+		}
+	}
+	
+	func invalidateTimer() {
+		schedulerTask?.cancel()
+		schedulerTask = nil
+	}
+}
--- a/Sources/EventProducer/Events/Logic/EventQueue.swift
+++ b/Sources/EventProducer/Events/Logic/EventQueue.swift
@@ -5,33 +5,34 @@ import SWXMLHash
 final class EventQueue {
 	static let databaseName = "EventQueueDatabase.sqlite"
 	
-	private var dbQueue: DatabaseQueue?
-
-	private func getDatabaseQueue() throws -> DatabaseQueue {
-		// if dbQueue is not nil, we have already set up the database
-		if let dbQueue {
-			return dbQueue
-		}
-
-		guard let databaseURL = FileManagerHelper.shared.eventQueueDatabaseURL else {
-			throw EventProducerError.eventQueueDatabaseURLFailure
-		}
-
-		do {
-			let databaseQueue = try DatabaseQueue(path: databaseURL.path)
-			try databaseQueue.write { db in
-				try db.create(table: EventPersistentObject.databaseTableName, ifNotExists: true) { table in
-					table.column(EventPersistentObject.columnID, .text).notNull()
-					table.column(EventPersistentObject.columnName, .text).notNull()
-					table.column(EventPersistentObject.columnConsentCategory, .text)
-					table.column(EventPersistentObject.columnHeaders, .text)
-					table.column(EventPersistentObject.columnPayload, .text)
-				}
+	private var _dbQueue: DatabaseQueue?
+	
+	private var dbQueue: DatabaseQueue {
+		get throws {
+			if let _dbQueue {
+				return _dbQueue
 			}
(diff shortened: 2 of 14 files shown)


Sources/EventProducer/TidalEventSender.swift
```diff
@@ -29,29 +29,36 @@ public final class TidalEventSender: EventSender {
...
 		self.config = config
-		self.scheduler = EventScheduler(consumerUri: config.consumerUri)
-		self.monitoringScheduler = MonitoringScheduler(consumerUri: config.consumerUri)
+		self.scheduler = .init(config: config, eventQueue: eventQueue, monitoring: monitoring)
+		self.monitoringScheduler = .init(config: config, eventQueue: eventQueue, monitoring: monitoring)
+
 		self.start()
 	}
 	
 	public func updateConfiguration(_ config: EventConfig) {
```
ksloginov, line 58: I'd delete the whole method

Sources/EventProducer/TidalEventSender.swift
```diff
@@ -29,29 +29,36 @@ public final class TidalEventSender: EventSender {
...
-		self.monitoringScheduler = MonitoringScheduler(consumerUri: config?.consumerUri)
+		self.eventQueue = EventQueue()
+		self.monitoringQueue = MonitoringQueue()
+		self.monitoring = Monitoring(monitoringQueue: monitoringQueue)
+		self.eventSubmitter = EventSubmitter(eventsQueue: eventQueue, monitoring: monitoring)
+		
 		self.fileManager = FileManagerHelper.shared
+		
+		self.scheduler = .init(config: config, eventQueue: eventQueue, monitoring: monitoring)
+		self.monitoringScheduler = .init(config: config, eventQueue: eventQueue, monitoring: monitoring)
```
ksloginov, line 47: At this point, to my understanding the config is `nil` and the `scheduler` won't do anything either way. Probably worth making scheduler optional (or `!`)? So that people don't forget to set config? 🤔

ksloginov (approved):
Awesome!

Added a couple of comments either way.

Sources/EventProducer/TidalEventSender.swift
e-kononenko: we can do something similar to `TidalAuth` implementation - throw an error if a config was not set (and schedulers were not initialized). 

But I wouldn't do it here, as it is not really related. Let's maybe make it in a separate PR?

ksloginov: Sure, if you prefer this way.

Sources/EventProducer/Common/Scheduler.swift
```diff
@@ -0,0 +1,37 @@
...
+		
+		schedulerTask = Task { [weak self, headerHelper] in
+			repeat {
+				guard let self else {
+					return
+				}
+				
+				let schedulingTimeInNanoseconds = UInt64(schedulingTime) * NSEC_PER_SEC
+				
+				try await Task.sleep(nanoseconds: schedulingTimeInNanoseconds)
```
asendra, line 27: In player we have a convenience method in a extension for sleeping in seconds
```
static func sleep(seconds: Double) async throws
```

Maybe we can move it to Common.. but can be done in a cleanup Pr later.

Sources/EventProducer/Common/Scheduler.swift
```diff
@@ -0,0 +1,37 @@
...
+	/// The Scheduler applies a best-effort approach to sending events to the TL Consumer, meaning that it continuously tries to send
+	/// events while the app is executing,
+	/// regardless of current internet conditions or if the app considers itself to be online or not. As long as the app is
+	/// executing, the Scheduler shall be running
+	func runScheduling(with headerHelper: HeaderHelper) {
+		guard schedulerTask == nil else {
+			return
+		}
+		
+		schedulerTask = Task { [weak self, headerHelper] in
```
asendra, line 19: Nice, you remembered to retain `headerHelper` here 💪

asendra (approved):
Very very nice job @e-kononenko! Just as I had in mind 🤩

Sources/EventProducer/TidalEventSender.swift
e-kononenko: agree, let's make it in the next PR

Sources/EventProducer/Common/Scheduler.swift
e-kononenko: great idea! I think it is worth to do it in a cleanup PR

emin-grbo (approved):
Great work @e-kononenko ! 🤞this solves a lot of the issues we had🙌

Sources/EventProducer/TidalEventSender.swift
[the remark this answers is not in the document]
e-kononenko: I am going to make a new PR, because this is not really related. There I will put all the things we discussed in these comments 🙂

Sources/EventProducer/Events/Logic/EventQueue.swift
```diff
@@ -4,33 +4,35 @@ import SWXMLHash
 
 final class EventQueue {
 	static let databaseName = "EventQueueDatabase.sqlite"
+	
+	private var _dbQueue: DatabaseQueue?
+	
+	private var dbQueue: DatabaseQueue {
+		get throws {
+			if let _dbQueue {
```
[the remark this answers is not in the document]
e-kononenko: thanks to @ksloginov for this suggestion! 🙂

PR #36, opened 2024-07-09
Event producer improvements

e-kononenko:
Addressing comments for [this PR](https://github.com/tidal-music/tidal-sdk-ios/pull/35)

* Removed `updateConfiguration(_ config:)` method and adjusted related tests
* Made schedulers optionals and removed their initialization with empty config. 
* Invalidate schedulers every time the config function is called. 
* Throw `EventProducerError.notConfigured` if `sendEvent` is called, but the event sender is not configured
* Made `start()` function private, as it is not mentioned in our documentation and used only inside `TidalEventProducer`. Calling it from the outside would start additional scheduling timers.
* Unify sleep functionality between `Player` and `EventProducer` - moved `Task.sleep(seconds:)` to the common module
* Fixed `testSendEventFailsWhenConfigurationNotCalled`


Sources/EventProducer/TidalEventSender.swift
```diff
@@ -42,19 +43,22 @@ public final class TidalEventSender: EventSender {
...
+		self.monitoringScheduler?.invalidateTimer()
+
 		self.scheduler = .init(config: config, eventQueue: eventQueue, monitoring: monitoring)
 		self.monitoringScheduler = .init(config: config, eventQueue: eventQueue, monitoring: monitoring)
 
 		self.start()
 	}
 	
+	@available(*, deprecated)
 	public func updateConfiguration(_ config: EventConfig) {
```
ksloginov, line 62: I still vote to kill it, as it's not bringing value in the current state (on the contrary, it breaks the logging), and its removal will be a breaking change at some point, so better to remote it sooner rather than later.

ksloginov (approved):

Sources/EventProducer/TidalEventSender.swift
e-kononenko: I also thought to remove it tbh, but I am not sure what to do with these tests:

https://github.com/tidal-music/tidal-sdk-ios/blob/9a6383212dda76570e3fa24ddb9435af3b758ff1/Tests/EventProducerTests/EventsTests/EventsTests.swift#L79-L108

https://github.com/tidal-music/tidal-sdk-ios/blob/9a6383212dda76570e3fa24ddb9435af3b758ff1/Tests/EventProducerTests/EventsTests/EventsTests.swift#L130-L161

https://github.com/tidal-music/tidal-sdk-ios/blob/9a6383212dda76570e3fa24ddb9435af3b758ff1/Tests/EventProducerTests/EventsTests/EventsTests.swift#L163-L191

They all call `updateConfiguration` function. I could for example, call `config` function again instead...

Or maybe you have another idea?

e-kononenko: @ksloginov still waiting for your input... 🙂

e-kononenko: Changed [here](https://github.com/tidal-music/tidal-sdk-ios/pull/36/commits/0817ec3da354cc04705cfb1c3f7ae7f9c8eb7304)

PR #38, opened 2024-07-11
[Auth] Fix Credentials upgrade scenario after clientId change

asendra:
Don't be fooled by the small amount of changes, this took half a day to debug  due to the recompilation times we have on the iOS client 😭 

## Context

As part of the deprecation project on the TIDAL client, we are updating the clientId. This triggers a credentials upgrade flow, which was causing a logout in the app. No bueno!

## Description

* After much spelunking, it was isolated to the fact that `CredentialsProvider.isUserLoggedIn` was returning `false`. The reason for this was that `credentials.level` was `.client` instead of `.user` after doing the token credentials upgrade. 
* This happened because the `credentials.userId` was `nil`. The reason for this was that the `UpgradeResponse` payload does not contain the `userId`, so we were overwriting the previous `userId` with `nil` when storing the new upgraded credentials.

This seems to be correct based on internal documentation about the **_oAuth2 flow for upgrading a token to a new client version_**

## Solution

I decided to remove the `userId` from the `UpgradeResponse` object, and instead use the userId value in the storedToken we are upgrading from.


e-kononenko (approved):

asendra:
Closing as it was fixed in backend

PR #39, opened 2024-07-22
Fix EventQueue.addEvent crash

e-kononenko:
We open database in `EventQueue` (i.e we create an instance of `DatabaseQueue`) if it was not done before. If it is already there, we just return it.

However, if multiple threads try to do it simultaneously, the `if let _dbQueue` is `false` for all of them and thus, we create multiple instances of the database. Since we override `_dbQueue` every time, we assume that at some point we try to call already deallocated instance - and there is a crash
Screenshot 2024-07-22 at 17 02 20

Changing class to actor should solve this data race by providing thread-safe access to `dbQueue` computed property - so we will really have only one instance, and further calls will just return already existing `_dbQueue`

| Before    | After |
| -------- | ------- |
|Screenshot 2024-07-22 at 17 14 16| Screenshot 2024-07-22 at 17 13 38|

### DIFF
--- a/Sources/EventProducer/Events/Logic/EventQueue.swift
+++ b/Sources/EventProducer/Events/Logic/EventQueue.swift
@@ -2,7 +2,7 @@ import Foundation
 import GRDB
 import SWXMLHash
 
-final class EventQueue {
+actor EventQueue {
 	static let databaseName = "EventQueueDatabase.sqlite"
 	
 	private var _dbQueue: DatabaseQueue?
--- a/Sources/EventProducer/Monitoring/MonitoringQueue.swift
+++ b/Sources/EventProducer/Monitoring/MonitoringQueue.swift
@@ -1,7 +1,7 @@
 import Foundation
 import GRDB
 
-public final class MonitoringQueue {
+actor MonitoringQueue {
 	static let databaseName = "MonitoringDatabase.sqlite"
 	private var _databaseQueue: DatabaseQueue?
 	


ksloginov (approved):
Very neat use of the modern syntax.

emin-grbo (approved):
If this resolves it completely it would be really cool.
Not familiar with any possible pitfalls only with the definition of it.
Is it the same as using `@MainActor` macro?

asendra:
> Is it the same as using `@MainActor` macro?

yes, but that is a specialization to always use the Main Queue as the actor context.. and we don't ever want that... at all. 🙂

asendra:
To clarify the reentrancy issue a bit, and why it does not apply here.
I understand the example might not seem very realistic. Check the internet if you want more 😀 

```
actor Database {
  var rows: Int = 0
  
  func addRow() async {
    let currentNumberOfRows = rows // <- This is thread-safe
    await expensiveCheck()
    // Because there has been an await before, this line is in the same thread, 
    // but we can't guarantee that currentNumberOfRows is valid anymore
    // Another addRow could have been executed while awaiting the expensiveCheck() in this context
    if currentNumberOfRows == 0 { 
      prepareTable()
    }
    
    addRow()
    rows += 1
  }
  
  private expensiveCheck() async {
    .....
  }
}
```

e-kononenko:
@emin-grbo 

> Is it the same as using `@MainActor` macro?

Kinda.

The macro indicates that some task is performed in the global [main actor](https://www.avanderlee.com/swift/mainactor-dispatch-main-thread/) which is associated with the main thread. But you can create your own [actors](https://www.avanderlee.com/swift/actors/) for a synchronized access to their data.

e-kononenko:
> where we read the `dbQueue` value repeatedly

I think it would be more optimal to use `deleteAll` method instead of iterating on delivered events and calling `deleteOne` on each of them

asendra:
> I think it would be more optimal to use `deleteAll` method instead of iterating on delivered events and calling `deleteOne` for each of them

If you mean the `deleteAll` method from GRDB that accepts an array of keys, then yes, that would be the obvious way of implementing it...

The method only deletes all that were delivered, so we can't delete ALL events in the dB.

e-kononenko:
> > I think it would be more optimal to use `deleteAll` method instead of iterating on delivered events and calling `deleteOne` for each of them
> 
> If you mean the `deleteAll` method from GRDB that accepts an array of keys, then yes, that would be the obvious way of implementing it...
> 
> The method only deletes all that were delivered, so we can't delete ALL events in the dB.

yes yes, sure. IIUC correctly, we can pass a collection of keys to it and it will delete only them 🙂 


Screenshot 2024-07-22 at 18 36 22

PR #42, opened 2024-07-25
Logging system and logging in Auth

e-kononenko:
This PR introduces a logging system in the SDK (`Common` module), according to this [document](https://docs.google.com/document/d/1qqofrUqjzC4lEicSV9tN2xh_wybGTbyThTYpg_IPPdY/edit?usp=sharing)

Also, it adds logging integration to the `Auth` module and logs cases described in this [document](https://docs.google.com/document/d/1HccRIohr3UPKkM1j_1eN3GzP2b0vTi4dZVlvMZGRQLg/edit?usp=sharing) (except a couple of cases which need to be discussed additionally and will be added later).

At the moment, all cases in `Auth` are logged with `.info` level (default), but we will change it later - it needs to be discussed additionally. 

More details:
* Logging system itself is quite simple - check `Logging`, `LogLevel` and `Loggable`
* `AuthLoggable` - all the cases we log in the `Auth` module
* Inject `Logger` to `Auth` via `AuthConfig` - the client will decide whether and how it wants to log different cases
* Call `logger?.log()` function where it is needed. It required a bit of refactoring sometimes, but this code is covered by unit tests, and they pass
* Improved printing errors:
    * Made `TidalError` conform to `LocalizedError` and print all necessary information
    * In some cases we lost the original `Error` when throwing a `NetworkError` - that's why we had just `NetworkError(0)` in our existing logs and therefore it was not really useful. I fixed this.
* One of the initializers of `Credentials` was marked as `throws`, but in fact, it didn't throw anything. I fixed this.


e-kononenko:
Closed, as in the current state it doesn't meet the existing standards for logging. Instead, we are going to utilize https://github.com/apple/swift-log instead.

PR #43, opened 2024-07-30
Playlog tests: Align skipToNext behavior with Android: don't do anything when there's no nextItem set up

ceciliasaraiva:
In order to fix the playlog tests, we need to address one of the different behaviors observed. So in this PR we address this different behavior - used by new playlog tests coming up in next PRs.

⚠️ The functionality itself (skip to next) is still under testing and therefore under control in case of any issues - the reason why I haven't added a FF.

---

This pull request includes changes to the `PlayerEngine` class in the `PlayerEngine.swift` file to streamline the `skipToNext` method and improve code readability.

Code simplification:

* [`Sources/Player/PlaybackEngine/PlayerEngine.swift`](diffhunk://#diff-e72feb354eac41dcce7baa31a6b2455945960a35c569f3ab640daa0e629a2015L184-R187): Simplified the `skipToNext` method by combining the assignment and the `guard` statement for `playerItem`, ensuring the current item is unloaded only when the next item is available.

### DIFF
--- a/Sources/Player/PlaybackEngine/PlayerEngine.swift
+++ b/Sources/Player/PlaybackEngine/PlayerEngine.swift
@@ -181,13 +181,10 @@ final class PlayerEngine {
 
 	func skipToNext(timestamp: UInt64) {
 		queue.dispatch {
-			let playerItem = self.nextItem
-			self.currentItem?.unloadFromPlayer()
-
-			guard let playerItem else {
-				self.reset(to: .IDLE)
+			guard let playerItem = self.nextItem else {
 				return
 			}
+			self.currentItem?.unloadFromPlayer()
 
 			// The order is important due to logic for emitting events in didSet of the items.
 			self.currentItem = playerItem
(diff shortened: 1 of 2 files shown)


asendra (commented):
Are we sure this is desired behaviour?

Before skip to next, the client would either load the next item into Player, or, if there is no next item, do something else I guess?

With this change, the next button will do nothing if there is no next track, right? (this can be tested if you disable `Autoplay` ang go to the last item in the queue)

Or is the button disabled, so there is no way for the client to call `skipToNext` in that case?

ceciliasaraiva:
> Are we sure this is desired behaviour?
> 
> Before skip to next, the client would either load the next item into Player, or, if there is no next item, do something else I guess?
> 
> With this change, the next button will do nothing if there is no next track, right? (this can be tested if you disable `Autoplay` ang go to the last item in the queue)
> 
> Or is the button disabled, so there is no way for the client to call `skipToNext` in that case?

This was a bug found during playlog tests based on their definition.

I believe it works the same way as if next is called when there is no next item. I suppose clients need to address if they want a different behavior, which I guess we don't, and that can be discussed separately.

emin-grbo (approved):
👍

asendra (approved):
As discussed, better to align on this, an this is a case that should not happen in our client because next button will be disabled.

PR #45, opened 2024-07-30
Playlog tests: Add tests for playlog - setNext, load, play and skipToNext

ceciliasaraiva:
After aligning the behavior of `skipToNext`, in this PR, I add a test (setNext, load, play and skipToNext) which would fail otherwise.

---

This pull request introduces a new test case to the `PlayLogTests` suite in `PlayLogTests.swift`. The test case verifies the behavior of the player engine when setting the next media product, loading, playing, and skipping to the next item.

### Additions to `PlayLogTests.swift`:

* Added `test_setNext_and_load_and_play_and_skipToNext` to verify the sequence of setting the next media product, loading, playing, and skipping to the next item. This includes setting up media products, loading the first media product, playing it, and asserting the play log events.


asendra (approved):
This is correct based on the previous change. Just need to make sure that it is correct (and how it is handled on the client)

PR #47, opened 2024-07-30
Playlog tests: Add missing tests

ceciliasaraiva:
Add missing tests for playlog: load, play, reset the player engine, set up a new player engine, load and play the same track and reset; load, play, reset the player engine, set up a new player engine, load and play another track and reset

---

This pull request enhances the `PlayLogTests` in `Tests/PlayerTests/Playlog/PlayLogTests.swift` by adding two new test functions. These tests cover scenarios involving loading, playing, resetting the player engine, and verifying play log events.

New test functions added:

* [`test_load_and_play_and_resetPlayerEngine_and_setUpANewPlayerEngine_and_load_and_play_another_track_and_reset`](diffhunk://#diff-ee8227a08a37793f9a69778a613ceea144edd7205bd0e71fb694ddbd89d34099R1203-R1371): This test covers the scenario of loading and playing a track, resetting the player engine, setting up a new player engine, loading and playing another track, and then resetting again. It verifies that the play log events are correctly recorded for both tracks.
* [`test_load_and_play_and_resetPlayerEngine_and_setUpANewPlayerEngine_and_load_and_play_same_track_and_reset`](diffhunk://#diff-ee8227a08a37793f9a69778a613ceea144edd7205bd0e71fb694ddbd89d34099R1203-R1371): This test covers the scenario of loading and playing a track, resetting the player engine, setting up a new player engine, loading and playing the same track again, and then resetting. It verifies that the play log events are correctly recorded for both instances of the same track.


emin-grbo (approved):
Would this remove flakiness completely or reduce it? Just wondering if we need to run it a bunch of times to verify how less of a flaky it is 😅

GJ!

ceciliasaraiva:
> Would this remove flakiness completely or reduce it? Just wondering if we need to run it a bunch of times to verify how less of a flaky it is 😅
> 
> GJ!

Still flaky unfortunately, I'm currently working on how to make them less flaky, and testing many many times locally.

asendra:
> Would this remove flakiness completely or reduce it? Just wondering if we need to run it a bunch of times to verify how less of a flaky it is 😅
> 
> GJ!

This is only adding new tests, so it can't be affecting the flakiness of previous tests in any way.

But they are not run in PR tests yet, so the flakiness is not affecting anyone.

asendra (approved):
This tests are ok, but I wonder if they are really testing different things.

`PlayerEngine` does not care if the `MediaProduct` is the same or a different one, so both tests from that point of view are testing exactly the same code.

Maybe the second test case was meant as a way to test the `.repeatOne` playback method? But that is a client implementation, it has nothing to do with the Player SDK in itself.

ceciliasaraiva:
> This tests are ok, but I wonder if they are really testing different things.
> 
> `PlayerEngine` does not care if the `MediaProduct` is the same or a different one, so both tests from that point of view are testing exactly the same code.
> 
> Maybe the second test case was meant as a way to test the `.repeatOne` playback method? But that is a client implementation, it has nothing to do with the Player SDK in itself.

Original was for the repeat since it exists in android, for ios it has less value. I guess for us is more to make sure the items get changed. If we would break that, many other tests would fail as well of course.

PR #48, opened 2024-07-30
[Auth] Logging system using swift-log

e-kononenko:
The previous [PR](https://github.com/tidal-music/tidal-sdk-ios/pull/42) because we decided to use a logging facade provided by [swift-log](https://github.com/apple/swift-log) library


Sources/Auth/Model/AuthLoggable.swift
```diff
@@ -0,0 +1,94 @@
...
+	case getCredentialsRefreshTokenIsNotAvailable
+	// swiftlint:enable identifier_name
+}
+
+// MARK: - Logging
+extension AuthLoggable {
+	private static let metadataErrorKey = "error"
+	private static let metadataReasonKey = "reason"
+	
+	var loggingMessage: Logger.Message {
```
ksloginov, line 26: I suspect, it could be simplified, as all the cases are just capitalized case descriptions? 🤔

ksloginov (approved):
LGTM

Sources/Auth/Model/AuthLoggable.swift
e-kononenko: let's think about it later :) just using `"\(self)"` duplicates the metadata and produces very long messages...

asendra (approved):
Lacking some context, but I'll give it a go.

I'm good with the choice of `swift-log` as the logging API, but I don't know what backend handler are we going to use on the client.

The best choice currently is `OSLog` I think, but `OSLog` it is recommended to not be used through the `swift-log` interface due to performance and customizability issues.

Also, I think we should have some kind of standards internally about the different log levels and when to use them. This PR uses `.trace` for everything on `Auth`? No idea if that is correct, but we can change later...

PR #64, opened 2024-08-27
Better logging in case of token refresh failures

e-kononenko:
* Include previous substatus - so we can understand what caused token refreshing
* In case of logout after token refresh failure - we log user (had refresh token) and client (did not have a refresh token) flows differently, so we can distinguish them later


asendra (approved):
At first I was a bit confused by the `previousSubstatus` naming, but It means the substatus that started the getCredentials/refreshToken flow, right?

e-kononenko:
> At first I was a bit confused by the `previousSubstatus` naming, but It means the substatus that started the getCredentials/refreshToken flow, right?

yes, the flow is:
any request -> 401 unauthorized with`previousSubstatus` -> refresh token -> error with `substatus`. We want to log both substatuses.

emin-grbo (approved):

PR #68, opened 2024-08-29
Added MediaProduct.extras from the Player spec

asendra:
New optional field in `MediaProduct` for the client to populate with metadata


ziad-halabi9 (approved):

andrelouw (approved):
LGTM 👍

Sources/Player/Common/Types/MediaProduct.swift
```diff
@@ -8,19 +8,22 @@ public class MediaProduct: Codable {
 	public let referenceId: String?
 	public let progressSource: Source?
 	public let playLogSource: Source?
+	public let extras: [String: String?]?
```
ceciliasaraiva, line 11: [question] just checking we want to have this one optional instead of having an empty array when there are no extras, then no need to handle optionals where needed.

ceciliasaraiva (approved):
just added one question regarding `extras` being optional.

Sources/Player/Common/Types/MediaProduct.swift
asendra: It's a Dict, not an Array, but that doesn't change your point :)

Ultimately, this is the same as it was implemented on Android. We had some discussions in the Android PR that you can check [here](https://github.com/tidal-music/tidal-sdk-android/pull/83)

This is an immutable property, so there are not many things to deal with due to the nullability I think? As a personal choice, I prefer it this way than sending an empty Dict, but If we had a set of standard practices for our APi, I would defer to that.

PR #72, opened 2024-09-02
Update unit-test.yml to fix broken release trigger

ceciliasaraiva:
Seems like I deleted one line by mistake in this [PR](https://github.com/tidal-music/tidal-sdk-ios/commit/7752931130586d7c8b51c0ae8a08f232e3f0c526). And from there on the trigger release got broken for 3 weeks. 🙀 

---

This pull request includes a small change to the `.github/workflows/unit-test.yml` file. The change adds a `workflow_call` trigger to the unit test workflow.

* [`.github/workflows/unit-test.yml`](diffhunk://#diff-d4223b44df678cece85f2542c19b9cd060d43b22f753c25214d5942ef2ca5d18R4): Added `workflow_call` trigger to the unit test workflow to enable reusability of the workflow in other workflows.

### DIFF
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -1,6 +1,7 @@
 name: Run Unit Tests
 
 on:
+  workflow_call:
   pull_request:
     branches:
       - '**'
--- a/.github/workflows/unit-test.yml
+++ b/.github/workflows/unit-test.yml
@@ -2,6 +2,7 @@ name: Run Unit Tests
 
 on:
   workflow_call:
+  workflow_dispatch:
   pull_request:
     branches:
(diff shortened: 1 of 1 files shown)


asendra (approved):
You also deleted `workflow_dispatch:` on that PR, which allowed us to run the workflow manually from the GitHub UI.

Are we sure we don't need that?

michpohl (approved):
👍

ceciliasaraiva:
> You also deleted `workflow_dispatch:` on that PR, which allowed us to run the workflow manually from the GitHub UI.
> 
> Are we sure we don't need that?

sure, can add it back as well 🤓

PR #76, opened 2024-09-04
[Player] Add database-backed persistence layer for Asset Cache

asendra:
Basic setup for the database-backed storage layer that will be used for the unified AssetCache with size control.

This PR defines an agnostic `CacheStorage` protocol that has regular CRUD operations returning `CacheEntry` objects, and an implementation of the protocol using `GRDB` to persist the data into an SQL database.

This implementation translates between the agnostic `CacheEntry` in the protocol definition and the database-specific DTO model `DBCacheEntryDTO` used to access the SQL database.

Besides the regular CRUD operations, it keeps track of the total size used and has a method to prune the cache up to a given maximum size.

There are also Unit Tests for the `DBCacheStorage` layer functionality.

## What's next

This cache storage layer will be connected in future PRs into the `GenericMediaPlayer` interface so that all player implementations share a single cache, allowing us to enforce size quotas.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -105,7 +105,7 @@ let package = Package(
 			dependencies: [
 				.common,
 				.KeychainAccess,
-				.Logging
+				.Logging,
 			],
 			plugins: [
 				.plugin(name: "SwiftLint", package: "SwiftLintPlugin"),
@@ -124,6 +124,7 @@ let package = Package(
 				.Kronos,
 				.auth,
 				.eventProducer,
+				.GRDB,
 			],
 			resources: [
 				.process("README.md"),
@@ -137,6 +138,7 @@ let package = Package(
 			dependencies: [
 				.player,
 				.auth,
+				.GRDB,
 			],
 			resources: [
 				.process("Resources"),
--- a/Sources/Player/PlaybackEngine/Internal/Cache/CacheStorage.swift
+++ b/Sources/Player/PlaybackEngine/Internal/Cache/CacheStorage.swift
@@ -20,4 +20,12 @@ protocol CacheStorage {
 	// MARK: - Get All CacheEntries
 
 	func getAll() throws -> [CacheEntry]
+
+	// MARK: - Calculate Total Size
+
+	func totalSize() throws -> Int
+
+	// MARK: - Prune to a Maximum Size
+
+	func pruneToSize(_ maxSize: Int) throws
 }
--- a/Sources/Player/PlaybackEngine/Internal/Cache/CacheStorage.swift
+++ b/Sources/Player/PlaybackEngine/Internal/Cache/CacheStorage.swift
@@ -1,0 +1,23 @@
+import Foundation
+
+protocol CacheStorage {
+	// MARK: - Save CacheEntry
+
+	func save(_ entry: CacheEntry) throws
+
+	// MARK: - Get CacheEntry by Key
+
+	func get(key: String) throws -> CacheEntry?
+
+	// MARK: - Delete CacheEntry by Key
+
+	func delete(key: String) throws
+
+	// MARK: - Update CacheEntry
+
+	func update(_ entry: CacheEntry) throws
+
+	// MARK: - Get All CacheEntries
+
+	func getAll() throws -> [CacheEntry]
+}
--- a/Sources/Player/PlaybackEngine/Internal/Cache/DBCacheEntryDTO.swift
+++ b/Sources/Player/PlaybackEngine/Internal/Cache/DBCacheEntryDTO.swift
@@ -1,0 +1,46 @@
+import Foundation
+import GRDB
+
+// MARK: - CacheEntryType + DatabaseValueConvertible
+
+extension CacheEntryType: DatabaseValueConvertible {}
+
+// MARK: - DBCacheEntryDTO
+
+struct DBCacheEntryDTO: Codable, FetchableRecord, PersistableRecord {
+	var key: String
+	var type: CacheEntryType
+	var url: URL
+	var lastAccessedAt: Date
+	var size: Int
+
+	static let databaseTableName = "cacheEntries"
+
+	enum Columns {
+		static let key = Column(CodingKeys.key)
+		static let type = Column(CodingKeys.type)
+		static let url = Column(CodingKeys.url)
+		static let lastAccessedAt = Column(CodingKeys.lastAccessedAt)
+		static let size = Column(CodingKeys.size)
+	}
+
+	init(key: String, type: CacheEntryType, url: URL, lastAccessedAt: Date = Date.now, size: Int) {
+		self.key = key
+		self.type = type
+		self.url = url
+		self.lastAccessedAt = lastAccessedAt
+		self.size = size
+	}
+
+	init(from entry: CacheEntry) {
+		key = entry.key
+		type = entry.type
+		url = entry.url
+		lastAccessedAt = entry.lastAccessedAt
+		size = entry.size
+	}
+
+	func toCacheEntry() -> CacheEntry {
+		CacheEntry(key: key, type: type, url: url, lastAccessedAt: lastAccessedAt, size: size)
+	}
+}
--- a/Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
+++ b/Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
@@ -1,0 +1,119 @@
+import Foundation
+import GRDB
+
+// MARK: - DBCacheStorage
+
+class DBCacheStorage {
+	private let dbQueue: DatabaseQueue
+
+	init(dbQueue: DatabaseQueue) {
+		self.dbQueue = dbQueue
+		try? initializedDatabase()
+	}
+
+	// MARK: - Calculate Total Size
+
+	func totalSize() throws -> Int {
+		try dbQueue.read { db in
+			try calculateTotalSize(db)
+		}
+	}
+
+	// MARK: - Prune to a Maximum Size
+
+	func pruneToSize(_ maxSize: Int) throws {
+		try dbQueue.write { db in
+			var currentSize = try self.calculateTotalSize(db)
+			guard currentSize > maxSize else {
+				return
+			}
+
+			let entries = try DBCacheEntryDTO
+				.order(DBCacheEntryDTO.Columns.lastAccessedAt.asc) // Oldest first
+				.fetchAll(db)
+
+			for entry in entries {
+				try entry.delete(db)
+				currentSize -= entry.size
+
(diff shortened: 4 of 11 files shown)


Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
```diff
@@ -0,0 +1,119 @@
+import Foundation
+import GRDB
+
+// MARK: - DBCacheStorage
+
+class DBCacheStorage {
```
bkdl, line 6: final?

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
```diff
@@ -0,0 +1,119 @@
...
+	func getAll() throws -> [CacheEntry] {
+		let entriesDTOs = try dbQueue.read { db in
+			try DBCacheEntryDTO.fetchAll(db)
+		}
+		return entriesDTOs.map { $0.toCacheEntry() }
+	}
+}
+
+private extension DBCacheStorage {
+	func initializedDatabase() throws {
```
bkdl, line 101: Tiny detail, but I'm guessing this should be named  as an action - `initializeDatabase`

Tests/PlayerTests/Player/PlaybackEngine/Internal/Cache/DBCache/DBCacheStorageTests.swift
```diff
@@ -0,0 +1,104 @@
+import GRDB
+@testable import Player
+import XCTest
+
+class DBCacheStorageTests: XCTestCase {
```
bkdl, line 5: final?

bkdl (approved):
Nice!

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
```diff
@@ -0,0 +1,119 @@
...
+						t.column("key", .text).primaryKey()
+						t.column("type", .text).notNull()
+						t.column("url", .text).notNull()
+						t.column("lastAccessedAt", .datetime).notNull()
+						t.column("size", .integer).notNull()
+					}
+				}
+			}
+		} catch {
+			print("Failed to initialize table \(DBCacheEntryDTO.databaseTableName): \(error)")
```
ceciliasaraiva, line 115: [suggestion] logging alert 🚨, feel free to add a todo for me - here and anywhere we want to log a message (I will check everything anyway, but this will guarantee I don't miss it by mistake) 🤓

Tests/PlayerTests/Player/PlaybackEngine/Internal/Cache/DBCache/DBCacheStorageTests.swift
```diff
@@ -0,0 +1,104 @@
...
+	var dbQueue: DatabaseQueue!
+	var cacheStorage: DBCacheStorage!
+
+	static let cacheEntry1 = CacheEntry.mock(
```
ceciliasaraiva, line 9: [suggestion] minor - I think these could be in a constants enum outside so we keep the class to only test/helper funcs and keep the actual class more concise, since most of the times these details are not relevant when reading the tests, just if we need to modify it 🙂

ceciliasaraiva (approved):
nice, I added some comments, let me know your thoughts.

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheEntryDTO.swift
```diff
@@ -0,0 +1,46 @@
+import Foundation
+import GRDB
+
+// MARK: - CacheEntryType + DatabaseValueConvertible
+
+extension CacheEntryType: DatabaseValueConvertible {}
+
+// MARK: - DBCacheEntryDTO
+
+struct DBCacheEntryDTO: Codable, FetchableRecord, PersistableRecord {
```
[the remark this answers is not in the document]
asendra: I was expecting the PR to kickstart a naming discussion :)

About the DB prefix, for me mainly I want to make it clear that it is the class for the GRDB implementation. there could be other like `UDCacheEntryDTO` for the UserDefaults one (which is how the current cache layer works)

I have also seen Entity used more for this, I had completely forgotten. Let's go with `DBCacheEntryEntity`?

Sources/Player/PlaybackEngine/Internal/Cache/CacheStorage.swift
```diff
@@ -0,0 +1,23 @@
+import Foundation
+
+protocol CacheStorage {
```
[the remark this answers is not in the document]
asendra: Do you mean a generic CRUD protocol for any entity? If so, that could work, but I prefer not to spend time right now on that, something we can do later when we have the sue cases more clear in the code, as a cleanup pass.

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
```diff
@@ -0,0 +1,119 @@
...
+
+	func totalSize() throws -> Int {
+		try dbQueue.read { db in
+			try calculateTotalSize(db)
+		}
+	}
+
+	// MARK: - Prune to a Maximum Size
+
+	func pruneToSize(_ maxSize: Int) throws {
```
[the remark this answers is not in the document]
asendra: The final usage is not in the PRs yet. It would mean making the `AVQueuePlayerWrapperLegacy` compatible with this new code which is a pain. So the integration is on hold until we cna get rid of the legacy player implementation. 

But the ideal will be that every time an item is added to the cache, this prunetoSize() will be called to enforce the maximum cache size.

But I will keep it in mind and think about the error handling a bit more.

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
asendra: got it 👍 
If the cache fails to initialize is not great, we should definitely log it, but the app can work without a cache.
Offliner is a different thing... if it fails to initialize we need to be careful how to handle that.

I will start a discussion later about if cache and offline should be different tables of the same DB, or different databases. (maybe to isolate issues between them)

Tests/PlayerTests/Player/PlaybackEngine/Internal/Cache/DBCache/DBCacheStorageTests.swift
asendra: yeah, agree. Because I only used them here I had not moved them out, but will do it.

Tests/PlayerTests/Player/PlaybackEngine/Internal/Cache/DBCache/DBCacheStorageTests.swift
asendra: yep, will do :+1

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
asendra: yes! Good catch. At first, I was returning an initialized instance but I refactored.
I might still change how to handle the database initialization, but for now, I will rename it.

Package.swift
```diff
@@ -105,7 +105,7 @@ let package = Package(
 			dependencies: [
 				.common,
 				.KeychainAccess,
-				.Logging
+				.Logging,
```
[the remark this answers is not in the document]
ceciliasaraiva: I've created a PR addressing current issues and will send communication so everyone gets this set up. 🙌

Sources/Player/PlaybackEngine/Internal/Cache/CacheStorage.swift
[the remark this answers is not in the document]
ceciliasaraiva: ok, from the first and second PRs, it seems there are already very similar needs, so for now looks good 🤓

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheEntryDTO.swift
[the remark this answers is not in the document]
ceciliasaraiva: well, developers and naming... 💜 sometimes 💔. 😅
I'll share my opinion and the train of thoughts to justify it. 🤓 Might be related to each developer experience in the end and comes down to personal preference. 😅

I feel like the "DB" would make more sense as a prefix of the "entity" keyword instead of the name of the struct, since it's there as CacheEntry entity in GRDB or in User Defaults - something like: "CacheEntryDBEntity" and "CacheEntryUDEntity". For me, they are all related to CacheEntry, and then the entity type is different, one located in GRDB and another in user defaults.
And maybe not popular opinion but I would even have "GRDB" and "UserDefaults" instead of only "DB" and "UD" since just those two letters don't immediately convey "GRDB" and "User Defaults" _to me_, so IMO a longer name in this case makes the reader immediately understand what it is instead of have questions about it before looking how it is used and concluding what it is.

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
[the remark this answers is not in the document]
ceciliasaraiva: Got it! Yes, it will be great to have it in mind since there are so many throws in each PR and therefore many points of failure.

Sources/Player/PlaybackEngine/Internal/Cache/DBCacheStorage.swift
ceciliasaraiva: Yes, do we know if we lose some time if we have different ones? I think the isolation you mentioned makes sense since there could be an issue in one and not the other, but there might be other factors to consider.

Tests/PlayerTests/Player/PlaybackEngine/Internal/Cache/DBCache/DBCacheStorageTests.swift
```diff
@@ -0,0 +1,104 @@
...
+
+		let fetchedEntry = try XCTUnwrap(cacheStorage.get(key: DBCacheStorageTests.cacheEntry1.key))
+		XCTAssertEqual(fetchedEntry, DBCacheStorageTests.cacheEntry1)
+	}
+
+	func testUpdateCacheEntry() throws {
+		try cacheStorage.save(DBCacheStorageTests.cacheEntry1)
+
+		// Update the entry
+		var updatedSize = 1000
```
ceciliasaraiva, line 54: in case you have missed it, there's this warning here which we can address

PR #77, opened 2024-09-05
[Player] Add database-backed persistance layer for Offline

asendra:
Basic setup for the database-backed storage layer that will be used for the Offline implementation.

This PR defines an agnostic `OfflineStorage` protocol that has regular CRUD operations returning `OfflineEntry` objects, and an implementation of the protocol using `GRDB` to persist the data into an SQL database.

This implementation translates between the agnostic `OfflineEntry` in the protocol definition and the database-specific DTO model `DBOfflineEntryDTO` used to access the SQL database.

The implementation provides all the regular CRUD operations.

There are also Unit Tests for the Storage layer functionality.

## What's next

This offline storage layer will be connected in future PRs into the `OfflineEngine` and `PlaybackEngine` layers.

### DIFF
--- a/Sources/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineEntryDTO.swift
+++ b/Sources/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineEntryDTO.swift
@@ -1,0 +1,104 @@
+import Foundation
+import GRDB
+
+// MARK: - DBOfflineEntryDTO
+
+struct DBOfflineEntryDTO: Codable, FetchableRecord, PersistableRecord {
+	let productId: String
+	let productType: ProductType
+	let assetPresentation: AssetPresentation
+	let audioMode: AudioMode?
+	let audioQuality: AudioQuality?
+	let audioCodec: AudioCodec?
+	let audioSampleRate: Int?
+	let audioBitDepth: Int?
+	let videoQuality: VideoQuality?
+	let revalidateAt: UInt64?
+	let expiry: UInt64?
+	let mediaType: String?
+	let albumReplayGain: Float?
+	let albumPeakAmplitude: Float?
+	let trackReplayGain: Float?
+	let trackPeakAmplitude: Float?
+	let mediaBookmark: Data?
+	let licenseBookmark: Data?
+
+	static let databaseTableName = "offlineEntries"
+
+	enum Columns {
+		static let productType = Column(CodingKeys.productType)
+		static let productId = Column(CodingKeys.productId)
+		static let assetPresentation = Column(CodingKeys.assetPresentation)
+	}
+
+	init(from entry: OfflineEntry) {
+		productId = entry.productId
+		productType = entry.productType
+		assetPresentation = entry.assetPresentation
+		audioMode = entry.audioMode
+		audioQuality = entry.audioQuality
+		audioCodec = entry.audioCodec
+		audioSampleRate = entry.audioSampleRate
+		audioBitDepth = entry.audioBitDepth
+		videoQuality = entry.videoQuality
+		revalidateAt = entry.revalidateAt
+		expiry = entry.expiry
+		mediaType = entry.mediaType
+		albumReplayGain = entry.albumReplayGain
+		albumPeakAmplitude = entry.albumPeakAmplitude
+		trackReplayGain = entry.trackReplayGain
+		trackPeakAmplitude = entry.trackPeakAmplitude
+		mediaBookmark = try? entry.mediaUrl?.bookmarkData()
+		licenseBookmark = try? entry.licenseUrl?.bookmarkData()
+	}
+
+	func toOfflineEntry() -> OfflineEntry {
+		OfflineEntry(
+			productId: productId,
+			productType: productType,
+			assetPresentation: assetPresentation,
+			audioMode: audioMode,
+			audioQuality: audioQuality,
+			audioCodec: audioCodec,
+			audioSampleRate: audioSampleRate,
+			audioBitDepth: audioBitDepth,
+			videoQuality: videoQuality,
+			revalidateAt: revalidateAt,
+			expiry: expiry,
+			mediaType: mediaType,
+			albumReplayGain: albumReplayGain,
+			albumPeakAmplitude: albumPeakAmplitude,
+			trackReplayGain: trackReplayGain,
+			trackPeakAmplitude: trackPeakAmplitude,
+			mediaUrl: mediaUrl(),
+			licenseUrl: licenseUrl()
+		)
+	}
+}
+
+private extension DBOfflineEntryDTO {
+	func mediaUrl() -> URL? {
+		var isStale = false
+		guard
+			let mediaBookmark,
+			let url = try? URL(resolvingBookmarkData: mediaBookmark, bookmarkDataIsStale: &isStale),
+			!isStale
+		else {
+			return nil
+		}
+		return url
+	}
+
+	func licenseUrl() -> URL? {
+		var isStale = false
+		guard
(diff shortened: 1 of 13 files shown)


Sources/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineEntryDTO.swift
```diff
@@ -0,0 +1,104 @@
+import Foundation
+import GRDB
+
+// MARK: - DBOfflineEntryDTO
+
+struct DBOfflineEntryDTO: Codable, FetchableRecord, PersistableRecord {
```
ceciliasaraiva, line 6: [suggestion] if we agree on a different naming from last PR, then let's apply it also for the entities here.

Sources/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+	// MARK: - Update OfflineEntry
+
+	func update(_ entry: OfflineEntry) throws {
+		let entryDTO = DBOfflineEntryDTO(from: entry)
+		try dbQueue.write { db in
+			try entryDTO.update(db)
+		}
+	}
+
+	// MARK: - Get All CacheEntries
```
ceciliasaraiva, line 53: [suggestion] I guess copy and paste issue
```suggestion
	// MARK: - Get All Offline Entries
```

Sources/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+
+	func getAll() throws -> [OfflineEntry] {
+		let entriesDTOs = try dbQueue.read { db in
+			try DBOfflineEntryDTO.fetchAll(db)
+		}
+		return entriesDTOs.map { $0.toOfflineEntry() }
+	}
+
+	// MARK: - clear All OfflineEntries
```
ceciliasaraiva, line 62: [suggestion] (not sure it can get more minor than this 😅, but since i notices I'm might as well write it) in case we add MARK with same name of func, then there's no "all":
```suggestion
	// MARK: - clear OfflineEntries
```

Sources/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+	}
+}
+
+private extension DBOfflineStorage {
+	func initializeDatabase() throws {
+		do {
+			try dbQueue.write { db in
+				if try !db.tableExists(DBOfflineEntryDTO.databaseTableName) {
+					try db.create(table: DBOfflineEntryDTO.databaseTableName) { t in
+						t.column("productId", .text).primaryKey()
```
ceciliasaraiva, line 77: [question] we can't access the coding keys from here, right? (would be to make a bit more safe).

Sources/Player/OfflineEngine/Internal/Storage/OfflineEntry.swift
```diff
@@ -0,0 +1,104 @@
...
+	var expiry: UInt64?
+	let mediaType: String?
+	let albumReplayGain: Float?
+	let albumPeakAmplitude: Float?
+	let trackReplayGain: Float?
+	let trackPeakAmplitude: Float?
+	var mediaUrl: URL?
+	var licenseUrl: URL?
+
+	var state: OfflineState {
```
ceciliasaraiva, line 24: [question] are these values same as in other platforms or do we keep different state for each platform?
[question 2] I see same `OFFLINED_BUT_NOT_VALID` and one `OFFLINED_AND_VALID`, so just checking indeed we want the same one in all conditions, or maybe we want to be more specific with the not valid condition, I think it can be good for future logging so we can more easily diagnose if it comes to it.

Sources/Player/OfflineEngine/Internal/Storage/OfflineEntry.swift
```diff
@@ -0,0 +1,104 @@
...
+	let videoQuality: VideoQuality?
+	var revalidateAt: UInt64?
+	var expiry: UInt64?
+	let mediaType: String?
+	let albumReplayGain: Float?
+	let albumPeakAmplitude: Float?
+	let trackReplayGain: Float?
+	let trackPeakAmplitude: Float?
+	var mediaUrl: URL?
+	var licenseUrl: URL?
```
ceciliasaraiva, line 22: [suggestion] minor, just noticed `Url` instead of `URL`, since it's an acronym:
```suggestion
	var mediaURL: URL?
	var licenseURL: URL?
```
If we rename it, there are more occurrences but I've just commented here.

Tests/PlayerTests/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineStorageTests.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import GRDB
+@testable import Player
+import XCTest
+
+class DBOfflineStorageTests: XCTestCase {
```
ceciliasaraiva, line 6:
```suggestion
final class DBOfflineStorageTests: XCTestCase {
```

Sources/Player/OfflineEngine/Internal/Storage/DBStorage/DBOfflineStorage.swift
```diff
@@ -0,0 +1,103 @@
...
+	func save(_ entry: OfflineEntry) throws {
+		let entryDTO = DBOfflineEntryDTO(from: entry)
+		try dbQueue.write { db in
+			try entryDTO.insert(db)
+		}
+	}
+
+	// MARK: - Get OfflineEntry by MediaProduct
+
+	func get(mediaProduct: MediaProduct) throws -> OfflineEntry? {
```
ceciliasaraiva, line 29: [question] to align with approach from previous PR and be more consistent, where we inform only the key, maybe here we do the same? Or change in the first one? Then it's either responsibility of this func to get the key or of the caller to inform it.
(I'm not adding comment, but it's also valid for the func below.)

ceciliasaraiva (approved):
Added some comments, some are basically the same as already pointed out in the previous PR.
Also there are some warnings of immutable values which we could also address.
I forgot to write in the previous PR: thanks for writing at least some very basic uts!

Sources/Player/OfflineEngine/Internal/Storage/OfflineEntry.swift
asendra: These are not coordinated with other platforms.

I could see us adding an extra state for `OFFLINED_BUT_EXPIRED` and `OFFLINED_BUT_NO_LICENSE`

asendra: Ok, I lied 😅 

This is part of the spec, so adding more would need us to sync with Android.
https://github.com/tidal-music/tidal-sdk/blob/main/Player.md#offlinestate

We could have an `InternalOfflineState` with more granularity, and map to OfflineState in the public API.

asendra: Added `InternalOfflineState` ✅

ceciliasaraiva: Oh nice, yes, maybe it doesn't make sense to expose all. Well, we can always get rid of it it proves to not be useful.

PR #78, opened 2024-09-05
[Player] Public interface for OfflineEngine

asendra:
Added the public interface for the `OfflineEngine` operations. This interface might change a little in the following Prs as we work to integrate this new implementation on the client.

### DIFF
--- a/Sources/Player/Player.swift
+++ b/Sources/Player/Player.swift
@@ -353,6 +353,26 @@ public extension Player {
 		}
 	}
 
+	func offline(mediaProduct: MediaProduct) -> Bool {
+		offlineEngine?.offline(mediaProduct: mediaProduct) ?? false
+	}
+
+	func deleteOffline(mediaProduct: MediaProduct) -> Bool {
+		offlineEngine?.deleteOffline(mediaProduct: mediaProduct) ?? false
+	}
+
+	func deleteAllOffline() -> Bool {
+		offlineEngine?.deleteAllOfflines() ?? false
+	}
+
+	func getOfflineState(mediaProduct: MediaProduct) -> OfflineState {
+		offlineEngine?.getOfflineState(mediaProduct: mediaProduct) ?? .NOT_OFFLINED
+	}
+
+	func setOfflinerDelegate(_ offlinerDelegate: OfflinerDelegate) {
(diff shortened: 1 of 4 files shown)


Sources/Player/Player.swift
```diff
@@ -360,6 +360,26 @@ public extension Player {
 		}
 	}
 
+	func offline(mediaProduct: MediaProduct) -> Bool {
+		offlineEngine?.offline(mediaProduct: mediaProduct) ?? false
+	}
+
+	func deleteOffline(mediaProduct: MediaProduct) -> Bool {
```
ceciliasaraiva, line 367: [suggestion] if we mean the media product, maybe it reads better
```suggestion
	func deleteOfflined(mediaProduct: MediaProduct) -> Bool {
```
or
```suggestion
	func deleteOfflinedMediaProduct(mediaProduct: MediaProduct) -> Bool {
```
or
```suggestion
	func deleteOfflinedItem(mediaProduct: MediaProduct) -> Bool {
```
or something similar.

Sources/Player/Player.swift
```diff
@@ -360,6 +360,26 @@ public extension Player {
 		}
 	}
 
+	func offline(mediaProduct: MediaProduct) -> Bool {
```
ceciliasaraiva, line 363: [suggestion] considering these are all public, maybe we can add some comments to each func.

Sources/Player/Player.swift
```diff
@@ -360,6 +360,26 @@ public extension Player {
...
+		offlineEngine?.offline(mediaProduct: mediaProduct) ?? false
+	}
+
+	func deleteOffline(mediaProduct: MediaProduct) -> Bool {
+		offlineEngine?.deleteOffline(mediaProduct: mediaProduct) ?? false
+	}
+
+	func deleteAllOffline() -> Bool {
```
ceciliasaraiva, line 371: [suggestion] I guess similar to above, just offline is not clear what it is offline, so maybe:
```suggestion
	func deleteAllOfflinedMediaProducts() -> Bool {
```
or
```suggestion
	func deleteAllOfflinedItems() -> Bool {
```
or something similar. I guess we make it consistent to how we name above.

ceciliasaraiva (approved):
nice, just added minor comments regarding naming and another one regarding docs.

PR #80, opened 2024-09-06
Do not set up logging system in auth

e-kononenko:
We're going to set it up (set a log handler) on clients instead, since it is a static property which may be called just once.

Instead, we introduce `isLoggingEnabled`, in case a client doesn't need logging

### DIFF
--- a/Sources/Auth/Model/AuthConfig.swift
+++ b/Sources/Auth/Model/AuthConfig.swift
@@ -1,7 +1,4 @@
 import Foundation
-import protocol Logging.LogHandler
-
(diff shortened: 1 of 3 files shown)


ceciliasaraiva (approved):
I feel that the `AuthLoggable` has more responsibilities now - know if it should log, actually log AND contains the types of log messages. IMO those should be more separate. For example, a logger, the types and in a configuration whether it should log.
Therefore my suggestion would be to have a little bit different structure to make it more solid 🥁. I’ll create a PR with what I came up with in the meantime to start the discussion on standardization of logging.
I'm approving so it doesn't block the update on the log handler we need to address.

PR #83, opened 2024-09-09
Format unformatted code

ceciliasaraiva:
It seems some people don't have the git pre-commit hook properly. 😢 As a result, we got a lot of code not following the agreed rules we have set up and sometimes this end up meaning we get the correct formatting in other people's PRs when they touch an unformatted file, which ends up polluting their PRs.
Therefore, in this PR, I run the formatter in whole code - apart from generated code.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -105,7 +105,7 @@ let package = Package(
 			dependencies: [
 				.common,
 				.KeychainAccess,
-				.Logging
+				.Logging,
 			],
 			plugins: [
 				.plugin(name: "SwiftLint", package: "SwiftLintPlugin"),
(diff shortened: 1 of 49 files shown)


emin-grbo (approved):
Superb! Lookin' good!

emin-grbo:
🧹

bkdl (approved):

asendra (approved):
Can we use the `.git-blame-ignore-revs` for this?

ceciliasaraiva:
> Can we use the `.git-blame-ignore-revs` for this?

Good idea. It's not set up here, I'll create a ticket for us and we can add it when it is done.

PR #85, opened 2024-09-11
Removing setupUserConfiguration from Player

ziad-halabi9:
- The user id is now fetched from the auth's `CredentialsProvider`. 
- The user client id is now supplied through a function block. 
- User's clientId is now made nullable.

### DIFF
--- a/Sources/Player/PlaybackEngine/Internal/Events/PlayerEventSender.swift
+++ b/Sources/Player/PlaybackEngine/Internal/Events/PlayerEventSender.swift
@@ -17,7 +17,7 @@ class PlayerEventSender {
 	private let featureFlagProvider: FeatureFlagProvider
 	private let eventSender: EventSender
 
-	@Atomic private var userConfiguration: UserConfiguration?
+	@Atomic private var userClientIdSupplier: (() -> Int)?
 
 	private let encoder: JSONEncoder
 	private let eventsDirectory: URL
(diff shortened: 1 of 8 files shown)


Sources/Player/PlaybackEngine/Internal/Events/PlayerEventSender.swift
```diff
@@ -209,11 +209,17 @@ private extension PlayerEventSender {
 				print("StreamingPrivilegesHandler failed to get credentials")
 				return
 			}
-
+			
+			let userClientId: Int? = if let userClientIdSupplier {
+				userClientIdSupplier()
+			} else {
+				nil
+			}
```
asendra, line 217: This could be cleaner like this

```suggestion
			let userClientId = userClientIdSupplier?()
```

Sources/Player/PlaybackEngine/Internal/Events/PlayerEventSender.swift
```diff
@@ -209,11 +209,17 @@ private extension PlayerEventSender {
...
+		
 			let user = User(
-				id: userConfiguration!.userId,
+				id: Int(userId!) ?? -1,
```
asendra, line 220: This causes a crash if `userId` is nil

ziad-halabi9: Yes I was aware of that. If you look up the code right before, we do return if the user is not authorized. So it should not happen. And if it does, then we have a deeper issue that we should solve. So I decided to force a crash in this case. Does that make sense? Or do you still prefer we set a fallback here?

asendra: yes, makes sense 👍 

Before we were also forcing a crash if we had an invalid userConfiguration so this is the same.

asendra (approved):

PR #86, opened 2024-09-11
Restructure logging to be more SOLID

ceciliasaraiva:
After discussion how to make logging system more SOLID, we have agreed on the first points: having a logger, which will be used by modules, and a loggable interface, which should be conformed to in the domain where we want to log. This guarantees some basic consistency in the logging usage.
- Add `TidalLogger`
- Add `TidalLoggable`
- Add `Logging` dependency in `Common`

### Next:
1. Update `Auth` to use the new API from this PR
2. Add logging in `Player`

---

This pull request introduces a logging framework to the project by adding a new `TidalLogger` struct and `TidalLoggable` protocol, along with necessary dependencies. These changes aim to standardize and simplify logging across the codebase.

### Logging Framework Integration:

* **Added `TidalLoggable` Protocol**: Introduced a protocol defining requirements for loggable objects, including properties for logging message, metadata, log level, and source. (`Sources/Common/Logger/TidalLoggable.swift`)
* **Added `TidalLogger` Struct**: Created a struct to provide a convenient API for logging, including methods to log messages at different levels and handle `TidalLoggable` objects. (`Sources/Common/Logger/TidalLogger.swift`)
* **Updated Package Dependencies**: Added the `Logging` dependency to the `Common` target in the `Package.swift` file to support the new logging framework. (`Package.swift`)

### DIFF
--- a/Sources/Common/Logger/TidalLogger.swift
+++ b/Sources/Common/Logger/TidalLogger.swift
@@ -11,17 +11,21 @@ public struct TidalLogger {
 }
 
 private extension TidalLogger {
-	func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata? = nil, source: String? = nil) {
-		self.logger.log(level: level, message, metadata: metadata, source: source)
+	func log(level: Logger.Level, message: Logger.Message, metadata: Logger.Metadata? = nil, source: String? = nil, file: String, function: String, line: UInt) {
+		self.logger.log(level: level, message, metadata: metadata, source: source, file: file, function: function, line: line)
 	}
 }
 
 public extension TidalLogger {
-	func log(loggable: TidalLoggable) {
+	func log(loggable: TidalLoggable, file: String = #file, function: String = #function, line: UInt = #line) {
 		self.log(
 			level: loggable.logLevel,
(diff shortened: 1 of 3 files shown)


Sources/Common/Logger/TidalLogger.swift
```diff
@@ -0,0 +1,27 @@
+import Logging
+
+/// Struct abstraction for a logger offering a convenient API for logging.
+public struct TidalLogger {
+	private var logger: Logger
+
+	public init(label: String, level: Logger.Level = .trace) {
```
asendra, line 7: swift-log documentation mentions `.trace` and `.debug` for library developers, so the default is good.

asendra (approved):
My only comment, for the future, not now, would be to have some kind of short guidelines on how much can we/should we log, what data is ok to log, when to use `.trace` vs `.debug` etc

ceciliasaraiva:
> My only comment, for the future, not now, would be to have some kind of short guidelines on how much can we/should we log, what data is ok to log, when to use `.trace` vs `.debug` etc

Totally, we do need guidelines. There are some recommended by Apple regarding levels, so potentially we could use that. But we still need to agree on this and general guidelines of usage.

PR #88, opened 2024-09-11
Fix parsing error in clientIdFromToken

ziad-halabi9:
Current the client id can result to null if the string is not a perfect base 64 encoded string. This PR fixes this issue.


asendra (approved):
I've noticed in `Player` we also do `Data(base64Encoded: manifest)` in `PlaybackInfoFetcher`... not sure if can be an issue there (hopefully we would have noticed!) but doesn't hurt to add it, right?

Also inside the Auth module, in `CodeChallengeBuilder` which was mostly code copied from Android, so maybe we need toa dd this on the iOS part of it.

ziad-halabi9:
@asendra Good things to point out.

For playback manifests, backend already sends them with padding. Hence we never encountered this issue. So we can keep it as it is.

For `CodeChallengeBuilder`, I'm not that familiar with its implementation. But eventually, that is appended to the login URL and used by backend. So they are doing the parsing. I would suggest leaving it as it is. Just giving @e-kononenko an fyi in case this gets brought up in the future.

PR #89, opened 2024-09-12
Add logging support in Player

ceciliasaraiva:
- Added logging support in Player through flag in initialization
- If logging is desired, then we initialize it and set it in PlayerWorld (to avoid passing to the many places where logging is needed)
- I've added all logging as error and using the place of the fail as the case name and name.

⚠️ For this logging names and messages, I've followed a pretty simple approach, using the name of the place, and not a big message. For the ones with error, we will send the error description. I've added all as errors. We can discuss one by one in this PR and decide the level, pick another level for all or leave as it is and update as we go and as we see it. We still need to have agreement on guidelines, so we can also postpone the decision after those are decided.

⚠️ For simplicity and consistency, I've followed same approach as in Auth, with a flag and the logger being instantiated inside the sdk. We may discuss if at any point it would make sense to inject a logger, giving the decision of the configuration of the logger to the user of the sdk.

⚠️ Already pointed out in the past, seems like the failing tests are not always reported. It was missing from last PR a fix for tests, and the CI didn't report failure.

⚠️ Another point is in which part of the flow the error is logged. In some cases, we get an error but create another one. For this first step, I've literally added logs in all levels of such flows - everywhere we had a catch. I'd prefer to add more now to start with, observe, and later get rid of what doesn't make sense. But if we already notice some places where it doesn't make sense, please point it out and we can already address those.

---

This pull request introduces extensive logging improvements across various modules to better handle and trace errors. The primary change is the addition of logging statements using the `PlayerWorld.logger` to capture errors in different parts of the codebase.

### Logging Enhancements:

* **Authentication Module:**
  * Added logging for failed attempts to get bearer token credentials and to convert credentials to a bearer token in `CredentialsProvider`. (`[Sources/Player/Common/Authentication/CredentialsProvider+Player.swiftR16-R21](diffhunk://#diff-bd58f9fbf2a2a8e53fdc8f4a92425e4517ce46457abbfde37bd28aa6cf959acaR16-R21)`)

* **DRM Handling:**
  * Added logging for failures in fetching licenses and processing content key requests in `FairPlayLicenseFetcher`, `StoredLicenseLoader`, and `StreamingLicenseLoader`. (`[[1]](diffhunk://#diff-5e6ac44115d32c3000091124c15de6b878eaadb027622d1963e8170163017037R116-R117)`, `[[2]](diffhunk://#diff-e4f018472818bcc484360d609feee78b8cbedcabfb950ee16a123acb1849060fR35)`, `[[3]](diffhunk://#diff-e4f018472818bcc484360d609feee78b8cbedcabfb950ee16a123acb1849060fR49)`, `[[4]](diffhunk://#diff-e34124f5ea781d42872d3e575597c3fb7d4367c795664d03401ee8c5a9b89f67R46)`)

* **HTTP Client:**
  * Added logging for payload encoding/decoding failures and request handling errors in `HttpClient` and `URLSession`. (`[[1]](diffhunk://#diff-789a68be9c0928fb27c866719c9ce8ccae0eccfefda14049d2b8264d42a8bc95R168)`, `[[2]](diffhunk://#diff-789a68be9c0928fb27c866719c9ce8ccae0eccfefda14049d2b8264d42a8bc95R177)`, `[[3]](diffhunk://#diff-789a68be9c0928fb27c866719c9ce8ccae0eccfefda14049d2b8264d42a8bc95R236)`)

* **Offline Engine:**
  * Added logging for download failures, license fetching failures, and item deletion errors in `DownloadTask`, `Downloader`, `LicenseDownloader`, and `OfflineEngine`. (`[[1]](diffhunk://#diff-b50645adbd7e57968139bc1048f2c4fa12db73e54e45e90d109373a1e08458fcR116)`, `[[2]](diffhunk://#diff-b50645adbd7e57968139bc1048f2c4fa12db73e54e45e90d109373a1e08458fcR137)`, `[[3]](diffhunk://#diff-14533cbb2dc8b962963699f85044a8dff91daa2a8ec925fd6c8d75df9ca42863R63)`, `[[4]](diffhunk://#diff-bc4cd2c6898affe73d605cdbf55cb93b1606ac51b1cb6dbf5effba04d8c636b9R43)`, `[[5]](diffhunk://#diff-bc4cd2c6898affe73d605cdbf55cb93b1606ac51b1cb6dbf5effba04d8c636b9R63)`, `[[6]](diffhunk://#diff-c25638cfdf9a5799584b59c57b4e374a03e64536da9e6dd5db37c53a33605ec8R98)`)

* **Playback Engine:**
  * Added logging for playback metadata reading failures in `AVQueuePlayerWrapper`. (`[Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swiftR462](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R462)`)

### New Logging Definitions:

* **New `PlayerLoggable` Enum:**
  * Introduced a comprehensive `PlayerLoggable` enum to define various loggable error cases for different modules. (`[Sources/Player/Common/Logging/PlayerLoggable.swiftR1-R347](diffhunk://#diff-241c41ad3dff9ee36736f5b5fc913320fd065b24ed9aeba7e1016bcf2345ef49R1-R347)`)

* **Integration with `PlayerWorldClient`:**
  * Added a `logger` property to `PlayerWorldClient` to facilitate centralized logging. (`[[1]](diffhunk://#diff-e93ff12ff899401df71646980605ab0bc4a489ffe43087ad2d8ff199515b9bb4R16)`, `[[2]](diffhunk://#diff-e93ff12ff899401df71646980605ab0bc4a489ffe43087ad2d8ff199515b9bb4L26-R28)`)*


Sources/Player/Common/Logging/PlayerLoggable.swift
```diff
@@ -0,0 +1,337 @@
...
+
+	var loggingMessage: Logger.Message {
+		return switch self {
+		// Event Sender
+		case .sendEventOfflinePlayFailed:
+			"EventSender-sendEventOfflinePlayFailed"
+		case .sendLegacyEventFailed:
+			"EventSender-sendLegacyEventFailed"
+		case .migrateLegacyDirectoryFailed:
+			"EventSender-migrateLegacyDirectoryFailed"
```
asendra, line 109: Can some of the name be generated automatically from the class name or something like that?

Sources/Player/Player.swift
```diff
@@ -109,14 +110,20 @@ public extension Player {
...
 		if shared != nil {
 			return nil
 		}
 
 		Time.initialise()
 
+		// When we have logging enabled, we create a logger to be used inside Player module and set it in PlayerWorld.
+		if shouldAddLogging {
+			PlayerWorld.logger = TidalLogger(label: "Player", level: .trace)
+		}
```
asendra, line 125: Also, is there any plan about how are we going to use this? 

- Always? 
- Controlled by a setting? 
- Controlled by a FeatureFlag?
- Enabled only in Beta builds?

asendra (approved):
This is outstanding @ceciliasaraiva!! It should help us quite a bit (as long as Firebase UI collaborates with us 😅 ) 

I don't see any issues with all of these, as they are all errors / exceptions, and it is not modifying the logic flow in any way.

We will need more planning and guidelines for other info, trace logs we might want to add, but I´m not sure how useful those would.. Do you know if we can see the history of logs for a given user / session at the same time? (to follow the trace)

Also, Do you have examples of the formatting and how they would appear on firebase?

Sources/Player/Common/Logging/PlayerLoggable.swift
ceciliasaraiva: I think we can get from the class name, I'll check that. As for the func, I'm not so sure. If we use #func, it returns the whole naming in the func and a specific format. Maybe with some formatting we could do it, to get it until the first parenthesis. 🤓 Do you mind if I add a ticket for this to be checked separately?

Sources/Player/Player.swift
ceciliasaraiva: The consumer of the sdk will decide this.

ceciliasaraiva:
> This is outstanding @ceciliasaraiva!! It should help us quite a bit (as long as Firebase UI collaborates with us 😅 )
> 
> I don't see any issues with all of these, as they are all errors / exceptions, and it is not modifying the logic flow in any way.
> 
> We will need more planning and guidelines for other info, trace logs we might want to add, but I´m not sure how useful those would.. Do you know if we can see the history of logs for a given user / session at the same time? (to follow the trace)
> 
> Also, Do you have examples of the formatting and how they would appear on firebase?

We can discuss this offline since it's responsibility of the consumer of the sdk. Regarding guidelines, we will come up with proposal and have a general agreement.

PR #93, opened 2024-09-16
[Player] Better OfflineStorage and OfflineEngine initialization

asendra:
While `OfflineEngine` is still under a development flag, both `OfflineEngine` and `OfflineStorage` should be optional, and the lack of any should not affect Player functionality.

This way we isolate better any possible issues with the new code.

I've also improved the database initialization for the `OfflineStorage` layer, by creating it in the Application Support directory, which follows the platform's recommended suggestions, and the database is inside a specific folder, making it easy to remove it or move it in the future, as the SQLite database creates multiple files for it.

### DIFF
--- a/Sources/Player/Common/Logging/PlayerLoggable.swift
+++ b/Sources/Player/Common/Logging/PlayerLoggable.swift
@@ -189,7 +189,7 @@ extension PlayerLoggable {
 		case .downloadFinalizeFailed:
 			"DownloadTask-downloadFinalizeFailed"
 
-		// Offline Engine
+		// GRDBOfflineStorage
 		case .withDefaultDatabase:
 			"GRDBOfflineStorage-withDefaultDatabase"
 
--- a/Sources/Player/Common/Logging/PlayerLoggable.swift
+++ b/Sources/Player/Common/Logging/PlayerLoggable.swift
@@ -2,13 +2,17 @@ import Common
 import Foundation
 import Logging
 
+// MARK: - Constants
+
 private enum Constants {
 	static let metadataErrorKey = "error"
 }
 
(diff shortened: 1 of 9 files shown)


Sources/Player/Common/Logging/PlayerLoggable.swift
```diff
@@ -2,13 +2,17 @@ import Common
 import Foundation
 import Logging
 
+// MARK: - Constants
```
ceciliasaraiva, line 5: weird, why my formatter formatted some things and not others 🤔 maybe I'll do the setup again just in case

ceciliasaraiva (approved):
nice! just two minor comments

ceciliasaraiva:
If we plan to include this change in 34, please push the appropriate changelog change to [this PR](https://github.com/tidal-music/tidal-sdk-ios/pull/94). Thanks!

Sources/Player/Player.swift
```diff
@@ -173,44 +167,53 @@ public extension Player {
...
-			networkMonitor,
-			playerEventSender,
-			notificationsHandler,
-			featureFlagProvider,
-			externalPlayers,
-			credentialsProvider
-		)
-
+		// For now, OfflineStorage and OfflineEngine are optional.
+		// Once the functionality is final it should not be, as Player can't work with a working OfflineEngine.
```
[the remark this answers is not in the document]
asendra: You are correct! Great catch

Sources/Player/Common/Logging/PlayerLoggable.swift
asendra: This is my setup
```
➜  ~ swiftlint --version
0.55.1
```

PR #95, opened 2024-09-18
Add SendAllEvents Method

emin-grbo:
Allow the client to send all events implicitly 🚀

### DIFF
--- a/Sources/EventProducer/TidalEventSender.swift
+++ b/Sources/EventProducer/TidalEventSender.swift
@@ -125,6 +125,7 @@ public final class TidalEventSender: EventSender {
 		do {
 			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
 		} catch let EventProducerError.unauthorized(code) {
+			print("⛔️\(#function): EventProducer is unauthorized with code: \(code)")
 			_ = try await headerHelper.credentialsProvider?.getCredentials(apiErrorSubStatus: code.description)
 			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
 		} catch {
--- a/Sources/EventProducer/TidalEventSender.swift
+++ b/Sources/EventProducer/TidalEventSender.swift
@@ -118,6 +118,11 @@ public final class TidalEventSender: EventSender {
 		)
 		endOutage(eventName: event.name)
 	}
+	
+	public func sendAllEvents() async throws {
+			let headerHelper = HeaderHelper(credentialsProvider: config?.credentialsProvider)
+			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
+	}
 
 	private func start() {
 		let headerHelper = HeaderHelper(credentialsProvider: config?.credentialsProvider)
--- a/Sources/EventProducer/TidalEventSender.swift
+++ b/Sources/EventProducer/TidalEventSender.swift
@@ -124,7 +124,7 @@ public final class TidalEventSender: EventSender {
 		let headerHelper = HeaderHelper(credentialsProvider: config?.credentialsProvider)
 		do {
 			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
-		} catch let EventProducerError.unauthorized(code) {
+		} catch let EventProducerError.unauthorized(code) where code <= Constants.httpUnauthorized {
 			print("⛔️\(#function): EventProducer is unauthorized with code: \(code)")
 			_ = try await headerHelper.credentialsProvider?.getCredentials(apiErrorSubStatus: code.description)
 			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
--- a/Sources/EventProducer/TidalEventSender.swift
+++ b/Sources/EventProducer/TidalEventSender.swift
@@ -119,9 +119,17 @@ public final class TidalEventSender: EventSender {
 		endOutage(eventName: event.name)
 	}
 	
+	/// Used to explicitly send all events directly from the scheduler. Usage would depend on the client (ex. when app moves to the background)
 	public func sendAllEvents() async throws {
 		let headerHelper = HeaderHelper(credentialsProvider: config?.credentialsProvider)
-		try await scheduler?.sendAllEvents(headerHelper: headerHelper)
+		do {
+			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
+		} catch let EventProducerError.unauthorized(code) {
+			_ = try await headerHelper.credentialsProvider?.getCredentials(apiErrorSubStatus: code.description)
+			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
+		} catch {
+			throw error
+		}
 	}
 
 	private func start() {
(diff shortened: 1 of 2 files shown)


Sources/EventProducer/TidalEventSender.swift
```diff
@@ -118,6 +118,11 @@ public final class TidalEventSender: EventSender {
 		)
 		endOutage(eventName: event.name)
 	}
+	
+	public func sendAllEvents() async throws {
+		let headerHelper = HeaderHelper(credentialsProvider: config?.credentialsProvider)
+		try await scheduler?.sendAllEvents(headerHelper: headerHelper)
+	}
```
asendra, line 125: Let's see what others think too, but should we try to recover form an expired token error, and retry?
I think auth errors should be handled internally.

```
 catch let EventProducerError.unauthorized(code) {
			_ = try await headerHelper.credentialsProvider?.getCredentials(apiErrorSubStatus: code.description)
			try await scheduler?.sendAllEvents(headerHelper: headerHelper)
} catch {
    throw error
}
```

asendra (commented):
This approach will work most of the time by itself, but we need to take some edge cases into consideration when using it on the client.

If the app is not playing, it will be suspended almost instantly, so the events might not get sent (specially on less than ideal network conditions)

There is a way to extend the cpu time a bit to finish some cleanup operations before getting suspended.
We can use the `beginBackgroundTask(expirationHandler:)` [API](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) for this. 

This can be done on the client side, not here, so there is no need to add it in this PR, but I wanted to share it here so we are all aware.

There is another API, [BGTaskScheduler](https://developer.apple.com/documentation/backgroundtasks/bgtaskscheduler), which is more advanced and complicated, that we could use to schedule sending the events at certain times while suspended, but we don't need it.

Sources/EventProducer/TidalEventSender.swift
asendra: cc @ceciliasaraiva @e-kononenko

emin-grbo: Although this is not too time critical, this is a great addition, love it! 🙌 great to handle this edge case 🚀

emin-grbo:
> This approach will work most of the time by itself, but we need to take some edge cases into consideration when using it on the client.
> 
> If the app is not playing, it will be suspended almost instantly, so the events might not get sent (specially on less than ideal network conditions)
> 
> There is a way to extend the cpu time a bit to finish some cleanup operations before getting suspended. We can use the `beginBackgroundTask(expirationHandler:)` [API](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) for this.
> 
> This can be done on the client side, not here, so there is no need to add it in this PR, but I wanted to share it here so we are all aware.
> 
> There is another API, [BGTaskScheduler](https://developer.apple.com/documentation/backgroundtasks/bgtaskscheduler), which is more advanced and complicated, that we could use to schedule sending the events at certain times while suspended, but we don't need it.

We should get 5seconds of background work which should be enough to handle any small batch of evens that did occur between 30seconds schedule
https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/extending_your_app_s_background_execution_time

We could look into extending that time on clien-side but maybe no need to make it more complex for something that is not too time critical, just "good to have"?

asendra:
> > This approach will work most of the time by itself, but we need to take some edge cases into consideration when using it on the client.
> > If the app is not playing, it will be suspended almost instantly, so the events might not get sent (specially on less than ideal network conditions)
> > There is a way to extend the cpu time a bit to finish some cleanup operations before getting suspended. We can use the `beginBackgroundTask(expirationHandler:)` [API](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) for this.
> > This can be done on the client side, not here, so there is no need to add it in this PR, but I wanted to share it here so we are all aware.
> > There is another API, [BGTaskScheduler](https://developer.apple.com/documentation/backgroundtasks/bgtaskscheduler), which is more advanced and complicated, that we could use to schedule sending the events at certain times while suspended, but we don't need it.
> 
> We should get 5seconds of background work which should be enough to handle any small batch of evens that did occur between 30seconds schedule [developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/extending_your_app_s_background_execution_time](https://developer.apple.com/documentation/uikit/app_and_environment/scenes/preparing_your_ui_to_run_in_the_background/extending_your_app_s_background_execution_time)
> 
> We could look into extending that time on clien-side but maybe no need to make it more complex for something that is not too time critical, just "good to have"?

Yeah, that would be enough, depending on network conditions.
Extending the time is very very easy on the client side, but I agree it is just a nice to have.

Sources/EventProducer/TidalEventSender.swift
ceciliasaraiva: I don't have context over all the auth error handling, it would be great at some point to have some consistent behavior in whole sdk. 🤓
As for this specific logic, would it depend on the code received to decide if we retry again or not? Otherwise, simply retrying it every time would also work. And maybe we actually log that error in the first catch from the suggestion - which we won't be aware of in case the retry succeeds.

emin-grbo: As user would already be aware of this error on client side (they have no auth), maybe we don't have to notify them ourselves, a client would do that in more evident manner.
Or did you mean to just log it to console for a more transparent approach?

Bare in mind, this will hopefully happen when the app is not being used at all anymore.

ceciliasaraiva: Hmm, I don't mean necessarily notifying. My question which was not clear, sorry for that, is that maybe our decision to directly retry or not could be decided upon code received from `EventProducerError.unauthorized(code)` - since I don't have whole context, I'm not sure if this makes sense or not, you will know more.
And then my other idea was to log it here for example:
```
 catch let EventProducerError.unauthorized(code) {
    // maybe we log here
    _ = try await headerHelper.credentialsProvider?.getCredentials(apiErrorSubStatus: code.description)
    try await scheduler?.sendAllEvents(headerHelper: headerHelper)
} catch {
    throw error
}
```

ceciliasaraiva (approved):
I've added a small suggestion regarding logging.

Sources/EventProducer/TidalEventSender.swift
emin-grbo: Oh I see. Might be interesting.
I can try and find out if there are defined codes for this that we can filter out, but I believe we just use that to forward the value and do not have a `specified` list we can lean on. Checking!

Logging wise, sure, sounds good!! 🙌

emin-grbo:
Props for noticing the code specific response @ceciliasaraiva !🙌 
This aligns us with Auth a lot more now 👌 
And @asendra for the initial retry check approach, code is humming now 🚀

But do lmk if you feel we can improve this further. (I will look into the client changes too @asendra , as you mentioned before)

emin-grbo:
As agreed after a short sync with @ceciliasaraiva , `Logging` will be extended in a separate PR to align fully with the Auth and Player.

PR #96, opened 2024-09-19
[Player] Add OfflinePrivilegeCheck to bootstrap process

asendra:
We want to be able to check inside `PlaybackEngine` if the user has permissions to play an offlined track.

### DIFF
--- a/Sources/Player/PlaybackEngine/Internal/PlayerItemLoader.swift
+++ b/Sources/Player/PlaybackEngine/Internal/PlayerItemLoader.swift
@@ -5,15 +5,18 @@ import Foundation
 
 final class PlayerItemLoader {
 	private let offlineStorage: OfflineStorage?
+	private let offlinePlaybackPrivilegeCheck: (() -> Bool)?
 	private let playbackInfoFetcher: PlaybackInfoFetcher
 	private var playerLoader: PlayerLoader
 
 	init(
 		with offlineStorage: OfflineStorage?,
+		_ offlinePlaybackPrivilegeCheck: (() -> Bool)?,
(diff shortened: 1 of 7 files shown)


Sources/Player/PlaybackEngine/Internal/PlayerItemLoader.swift
```diff
@@ -42,7 +49,8 @@ private extension PlayerItemLoader {
 
 		if PlayerWorld.developmentFeatureFlagProvider.isOffliningEnabled,
 		   let offlineEntry = try? offlineStorage?.get(key: mediaProduct.productId),
-		   let offlinedMediaProduct = PlayableOfflinedMediaProduct(from: offlineEntry)
+		   let offlinedMediaProduct = PlayableOfflinedMediaProduct(from: offlineEntry),
+		   offlinePlaybackPrivilegeCheck?() ?? true == true
```
ceciliasaraiva, line 53: [suggestion] tiny suggestion to maybe store the result `offlinePlaybackPrivilegeCheck?() ?? true` to later compare in the if condition, i think it could improve slightly the readability of the condition.

ceciliasaraiva (approved):
super small suggestion

Sources/Player/PlaybackEngine/Internal/PlayerItemLoader.swift
asendra: yeah agree. 

I made this PR ready so I could check the linter issue on an old PR, but I want to make some more changes to it still

ceciliasaraiva: ahh, sorry about it, just let me know when ready to be checked

PR #97, opened 2024-09-19
Removed recent SwiftLint warnings

asendra:
As the title says.

[One of the warnings](https://realm.github.io/SwiftLint/non_optional_string_data_conversion.html) was related to how we were doing Data <-> String conversions, because if we use UTF8 encoding, it can never fail, so we can change to a non-optional way of doing it.

[More info here](https://github.com/realm/SwiftLint/issues/5263) and [here](https://www.swiftbysundell.com/tips/converting-between-string-and-data-without-optionals/)


emin-grbo (approved):
Had no idea, this is neat! 🙌

ceciliasaraiva (approved):
thanks!

PR #100, opened 2024-09-27
[Player] Upgraded OfflineEngine from dev flag to feature flag

asendra:
## Context
During development we had OfflineEngine functionality behind a development flag that was not controllable externally.

Now that the functionality is coming together and being integrated into the client, we want to control it through our FF system.


Sources/Player/PlaybackEngine/Internal/PlayerItemLoader.swift
```diff
@@ -40,8 +44,7 @@ private extension PlayerItemLoader {
 			return try await (metadata(of: storedMediaProduct), playerLoader.load(storedMediaProduct))
 		}
 
-		if PlayerWorld.developmentFeatureFlagProvider.isOffliningEnabled,
-		   let offlineEntry = try? offlineStorage?.get(key: mediaProduct.productId),
+		if let offlineEntry = try? offlineStorage?.get(key: mediaProduct.productId),
```
asendra, line 47: I think we don't need to control the flag here because storage will be nil when the flag is off.

ceciliasaraiva: Do we need to worry about scenario where user had the flag on (and had storage) and while using the app, it changed to off?

ceciliasaraiva (approved):
just added one question which I might be missing from whole flow

Sources/Player/PlaybackEngine/Internal/PlayerItemLoader.swift
asendra: I finally addressed this. Makes it a bit more complex, but I think it is worth it to avoid weird situations. Once the Flag is removed we will be able to simplify it.

Tests/PlayerTests/FeatureFlagProviderTests.swift
```diff
@@ -4,40 +4,70 @@ import EventProducer
...
 			eventSender: EventSenderMock()
 		)
 
 		let player = try XCTUnwrap(playerInstance)
 		XCTAssertNotNil(player.playerEngine)
+		XCTAssertNotNil(player.offlineStorage)
+		XCTAssertNotNil(player.offlineEngine)
+	}
+
+	func testOffliningIsInitializedAfterBootstrap() throws {
```
ceciliasaraiva, line 53: [suggestion] is we start with false, then would it make sense to assert nil before we change? And same if going from false to true

Sources/Player/PlaybackEngine/Internal/PlayerItemLoader.swift
ceciliasaraiva: nice, thanks!

Tests/PlayerTests/FeatureFlagProviderTests.swift
asendra: In my mind, that case is already covered in the test `func testOffliningIsNotInitialized() throws`

I guess both could be combined.

PR #101, opened 2024-09-27
[Player] Calculate size of downloaded MediaProducts

asendra:
## Context
As part of the new `OfflineEngine`, we want to calculate the total size of the downloaded Media so we can expose it to our users.

## Description
This PRs adds functionality to calculate both the size of progressive files downloaded, as well as the size of HLS streams downloaded.

HLS streams are stored as a folder with a bunch of files and subfolders, and we need to calculate the total size of all of those files.

### DIFF
--- a/Sources/Player/OfflineEngine/Internal/MediaDownloader.swift
+++ b/Sources/Player/OfflineEngine/Internal/MediaDownloader.swift
@@ -62,18 +62,15 @@ private extension MediaDownloader {
 	}
 
 	func calculateHLSStreamSize(for path: URL) -> UInt64 {
-		let fileManager = FileManager.default
 		var totalSize: UInt64 = 0
 
-		if let enumerator = fileManager.enumerator(
+		if let enumerator = PlayerWorld.fileManagerClient.enumerator(
 			at: path,
-			includingPropertiesForKeys: [.fileSizeKey],
-			options: [],
-			errorHandler: nil
+			includingPropertiesForKeys: [URLResourceKey.fileSizeKey]
 		) {
 			for case let fileURL as URL in enumerator {
 				do {
-					let fileAttributes = try fileURL.resourceValues(forKeys: [.fileSizeKey])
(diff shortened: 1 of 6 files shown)


Sources/Player/OfflineEngine/Internal/MediaDownloader.swift
```diff
@@ -60,6 +60,43 @@ private extension MediaDownloader {
 	func createTask(for url: URL) -> URLSessionDownloadTask {
 		urlDownloadSession.downloadTask(with: url)
 	}
+
+	func calculateHLSStreamSize(for path: URL) -> UInt64 {
+		let fileManager = FileManager.default
```
ceciliasaraiva, line 65: [suggestion] I think we can use `FileManagerClient`, we would need to add the `enumerator` there.

Sources/Player/OfflineEngine/Internal/MediaDownloader.swift
```diff
@@ -60,6 +60,43 @@ private extension MediaDownloader {
...
+			errorHandler: nil
+		) {
+			for case let fileURL as URL in enumerator {
+				do {
+					let fileAttributes = try fileURL.resourceValues(forKeys: [.fileSizeKey])
+					if let fileSize = fileAttributes.fileSize {
+						totalSize += UInt64(fileSize)
+					}
+				} catch {
+					print("Error reading file size: \(error.localizedDescription)")
```
ceciliasaraiva, line 81: [question] do we need proper log? I guess it depends on how this can this affect the offlining. (same for below)

ceciliasaraiva (approved):
nice, just two comments

Sources/Player/OfflineEngine/Internal/MediaDownloader.swift
asendra: Addressed 👍

Sources/Player/OfflineEngine/Internal/MediaDownloader.swift
asendra: Done!

PR #102, opened 2024-09-27
Simple logging API

e-kononenko:
There is no need to create `TidalLoggable` in simple scenarios.

This function allows to log without it.

Also, I made a `source` mandatory

### DIFF
--- a/Sources/Common/Logger/TidalLogger.swift
+++ b/Sources/Common/Logger/TidalLogger.swift
@@ -33,21 +33,18 @@ public extension TidalLogger {
 		message: String,
 		source: String,
 		level: Logger.Level? = nil,
-		metadata: [String: String]? = nil,
+		metadata: Logger.Metadata? = nil,
 		file: String = #fileID,
 		function: String = #function,
 		line: UInt = #line
 	) {
 		let level = level ?? logger.logLevel
 		let loggerMessage = Logger.Message(stringLiteral: message)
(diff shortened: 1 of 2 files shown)


Sources/Common/Logger/TidalLogger.swift
```diff
@@ -28,4 +28,30 @@ public extension TidalLogger {
 			line: line
 		)
 	}
+	
+	func log(
+		message: String,
+		source: String,
+		level: Logger.Level? = nil,
+		metadata: [String: String]? = nil,
```
ceciliasaraiva, line 36: [question] shall we consider keeping `Logger.Metadata` type here? So even for the simple cases, people can add other types apart from `String`?

ceciliasaraiva (approved):
nice one, thanks for adding this! I just added one question

Sources/Common/Logger/TidalLogger.swift
e-kononenko: Well, I thought to create 2 methods: with `Logger.Metadata` and with `[String: String]` (which will be the majority of the cases in my opinion), but then had problems that Swift couldn't select which function to use 🙂 

I will try to come up with something on Monday, if not - I'll just use `Logger.Metadata` 🙂

ceciliasaraiva: I think this works fine! 🙌 Once we have more usages also, we can adapt and make appropriate changes if this is not preferred by users of the logging.

PR #103, opened 2024-09-30
Better error descriptions

e-kononenko:
Do not use localized description for errors. Instead, use `String(describing:)` and `CustomStringConvertible`


ceciliasaraiva (approved):
nice one!

PR #105, opened 2024-10-04
As an attempt to fix a seek issue, add a FF which means we will pause and play around the seek call

ceciliasaraiva:
Attempt to address an issue we have around seeking.

⚠️ I'm testing our metrics to make sure nothing gets broken because of this, so the code is not final. Also, because our playlog tests are flaky, we cannot fully rely on them, so I'm adapting and testing locally since they are less flaky than in CI.

---

This pull request introduces a new feature flag for pausing and playing around seek operations in the player, along with the necessary logic to handle this behavior in the `AVQueuePlayerWrapper` and `AVQueuePlayerWrapperLegacy` classes. The most important changes are as follows:

### Feature Flags Update:

* [`Sources/Player/Common/FeatureFlags/FeatureFlagProvider.swift`](diffhunk://#diff-4cd187606f8aae83fb59b73a080a87896fe6033ce7fed43af1bc02b0d4b5dfc5R6-R19): Added `shouldPauseAndPlayAroundSeek` to the `FeatureFlagProvider` struct and its initializer.
* [`Sources/Player/Common/FeatureFlags/FeatureFlagProvider+Standard.swift`](diffhunk://#diff-357b6e003c6e1c07f64bda0a4006d43ba8928ecc38a2c1b7e8c5848e6f6b9fecL8-R9): Updated the standard feature flag provider to include `shouldPauseAndPlayAroundSeek`.
* [`Sources/Player/Mocks/Common/FeatureFlags/FeatureFlagProvider+Mock.swift`](diffhunk://#diff-29ec55675dc54daf5487ce14e6350480fed65a1f0163e4d1bda67fe83128c6deL8-R9): Updated the mock feature flag provider to include `shouldPauseAndPlayAroundSeek`.

### Player Wrapper Logic:

* `Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift`: 
  * Introduced `isSeeking` property and `shouldPauseAndPlayAroundSeek` computed property. [[1]](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R23) [[2]](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R38-R41)
  * Modified seek logic to pause and play around seek if the new feature flag is enabled. [[1]](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R192-R206) [[2]](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R216)
  * Updated `paused` method to check `isSeeking` status.

* `Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift`: 
  * Introduced `isSeeking` property and `shouldPauseAndPlayAroundSeek` computed property.
  * Modified seek logic to pause and play around seek if the new feature flag is enabled.
  * Updated `paused` method to check `isSeeking` status.

### DIFF
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
@@ -203,6 +203,7 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
 					return
 				}
 
+				asset.setAssetPosition(currentItem)
 			} else {
 				let completed = await self.player.seek(to: time)
 				guard completed, currentItem == self.player.currentItem, self.player.timeControlStatus == .playing else {
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
@@ -20,6 +20,7 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
 
 	@Atomic private var player: AVQueuePlayer
 	private var playerMonitor: AVPlayerMonitor?
+	private var isSeeking = false
 
 	private let contentKeyDelegateQueue: DispatchQueue = DispatchQueue(label: "com.tidal.player.contentkeydelegate.queue")
 	private let playbackTimeProgressQueue: DispatchQueue = DispatchQueue(label: "com.tidal.player.playbacktimeprogress.queue")
@@ -32,6 +33,10 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
 
 	private var isContentCachingEnabled: Bool {
 		featureFlagProvider.isContentCachingEnabled()
+	}
+
+	private var shouldPauseAndPlayAroundSeek: Bool {
+		featureFlagProvider.shouldPauseAndPlayAroundSeek()
 	}
 
 	private let supportedCodecs: [PlayerAudioCodec] = [
@@ -184,13 +189,29 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
 
 			self.delegates.seeking(in: asset)
 
-			let completed = await self.player.seek(to: time)
-			guard completed, currentItem == self.player.currentItem, self.player.timeControlStatus == .playing else {
-				return
-			}
+			if self.shouldPauseAndPlayAroundSeek {
+				if self.player.timeControlStatus == .playing {
+					self.isSeeking = true
+					await self.player.pause()
+				}
 
-			asset.setAssetPosition(currentItem)
-			self.delegates.playing(asset: asset)
+				let completed = await self.player.seek(to: time)
+				self.isSeeking = false
+				await self.player.play()
+
+				guard completed, currentItem == self.player.currentItem else {
+					return
+				}
+
+			} else {
+				let completed = await self.player.seek(to: time)
+				guard completed, currentItem == self.player.currentItem, self.player.timeControlStatus == .playing else {
+					return
+				}
+
+				asset.setAssetPosition(currentItem)
+				self.delegates.playing(asset: asset)
+			}
 		}
 	}
 
@@ -512,6 +533,9 @@ private extension AVQueuePlayerWrapper {
 	}
 
 	func paused(playerItem: AVPlayerItem) {
+		guard !isSeeking else {
+			return
+		}
 		queue.dispatch {
 			guard let asset = self.playerItemAssets[playerItem] else {
 				return
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
@@ -174,10 +174,9 @@ final class AVQueuePlayerWrapperLegacy: GenericMediaPlayer {
 					return
 				}
 			} else {
-				self.player.seek(to: seekTime) { completed in
-					guard completed, currentItem == self.player.currentItem, self.player.timeControlStatus == .playing else {
-						return
-					}
+				let completed = await self.player.seek(to: time)
+				guard completed, currentItem == self.player.currentItem, self.player.timeControlStatus == .playing else {
+					return
 				}
 			}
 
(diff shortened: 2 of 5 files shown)


Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
```diff
@@ -154,9 +159,26 @@ final class AVQueuePlayerWrapperLegacy: GenericMediaPlayer {
...
+
+				guard completed, currentItem == self.player.currentItem else {
+					return
+				}
+			} else {
+				self.player.seek(to: seekTime) { completed in
+					guard completed, currentItem == self.player.currentItem, self.player.timeControlStatus == .playing else {
+						return
+					}
+				}
```
asendra, line 181: Did you forget to return to the `await seek` version here?

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
```diff
@@ -154,9 +159,26 @@ final class AVQueuePlayerWrapperLegacy: GenericMediaPlayer {
...
+			let seekTime = CMTime(seconds: time, preferredTimescale: 1000)
+			if self.shouldPauseAndPlayAroundSeek {
+				if self.player.timeControlStatus == .playing {
+					self.isSeeking = true
+					await self.player.pause()
+				}
+
+				let completed = await self.player.seek(to: time)
+				self.isSeeking = false
+				await self.player.play()
```
asendra, line 171: if seek failed and completed is false, should we do play here?

asendra: Also, later we do 
```
self.delegates.playing(asset: asset)
```

But I think that should not be done in the case where we manually call play (It will do that already, so now we repeat the delegate callback)

asendra: (This applies to both Player implementations)

asendra (commented):
A couple of comments about some details, but I'm ok with this change if it fixes the issue. 👍

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
ceciliasaraiva: oh yes, I changed back after checking that and didn't commit, it is in staging, thanks for spotting!

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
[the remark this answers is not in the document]
ceciliasaraiva: I don't know. For my testing of the issue, it is indeed only in the end but I don't know if that is for all devices/versions, etc - if that matters at all. One thing I see as benefit is if we have any other issue (past/future), this could be the key to address it, and since it's behind the FF, we can be sure it doesn't create new issues and easily constraint a bit more if it does (or if we find other issues during wider testing). So even though I point that out, I wouldn't say I'm strong about it due to the limited testing for now (only my device, so could be as well missing some scenarios), therefore I can change it if you are strong about this suggestion.

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
ceciliasaraiva: Yes, I was not sure about the failed scenario. Let me do some testing around the unhappy scenario.

ceciliasaraiva: I tried to test to get actual completed false - I could just force it, but this doesn't apply to the actual player returning false since I would guess the internal state would be different.
I think if it fails/completed is false, we should still try to play it. This I believe would have the same behavior as current scenario when we don’t do pause/play, so it would follow the same existing approach.
As for the `self.delegates.playing(asset: asset)`, indeed there's no need to call it. I did remove it at some point - just saw I removed in the non legacy version but missed adding the asset position, so in the end both needed adjustments after I rearranged the code after testing.

ceciliasaraiva: removed duplicate playing call, and adjusted both players for the asset position.

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
```diff
@@ -184,13 +189,30 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
...
+				if self.player.timeControlStatus == .playing {
+					self.isSeeking = true
+					await self.player.pause()
+				}
 
-			asset.setAssetPosition(currentItem)
-			self.delegates.playing(asset: asset)
+				let completed = await self.player.seek(to: time)
+				self.isSeeking = false
+				await self.player.play()
```
asendra, line 200: From reading the documentation, seek should only fail if it is interrupted by other seek or similar operation.

When interrupted, the first seek finishes with completed = false and the next seek will finish with true.

In our case, because we have all the AVPlayer operations synchronous with the queue, I don't think it should return false.. so I think handling it like this is fine.

asendra (approved):

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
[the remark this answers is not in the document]
asendra: No, I'm ok with doing it for every seek, not only at the end. 
As you correctly point out, we have the flag as a safety measure.

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
ceciliasaraiva: yes, for testing I did the two seeks, but this scenario shouldn't ever happen so it was not "real", so I don't know how a real scenario would actually change the internal state which could - I don't know - possibly affect our current conditions.

PR #109, opened 2024-10-14
AnyCodable Extras

andrelouw:
## Context
This PR aims to introduce `AnyCodable` extras for the `MediaProduct` class, which will allow any aCodablea object from the client app to be passed as `extras`. 

## Description
- The [`AnyCodable`](https://github.com/Flight-School/AnyCodable) library was recently added to this project as part of the OpenAPI work being done. This simplified this piece of work quite a bit, as this library is exactly the solution that was proposed for this solution. 
- A new type (typealias for now) `AnyCodableDictionary = [String: AnyCodable?]` was introduced in the `Common` module. This was done to prevent the undesired spread of the `AnyCodable` type to the `Player` module. By doing this, the `Player` doesn't require new explicit dependencies at least. 
- In addition all `Encodable` types were extended with a `asAnyCodableDictionary()` adapter method to make it easy to convert any `Encodable` type to `AnyCodableDictionary`.
- A new type `MediaProduct.Extras` (`typealias` for `AnyCodableDictionary`  were introduced to further curb the spread of the new type for users of the `Player` module. With this type, users of the `Player` doesn't require any other imports other than the `Player`. This allows independent change and evolution of this type, without impacting users of the module too much.
- `MediaProduct.Extras` was also updated with its own `asMediaProductExtras()` conversion on `Encodable` types, which just uses `asAnyCodableDictionary()` under to hood. This is again to prevent the undesired spread of the `AnyCodable` type. 
- `PlayerEvent.Extras` was also updated accordingly with the same idea in mind. 
- Lastly, `AnyCodableDictionaryConversionTests` tests were added to prove the conversion to `AnyCodableDictionary` for any `Encodable` type.

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -71,7 +71,7 @@ let package = Package(
 		.target(
 			name: "Common",
 			dependencies: [
-				.Logging
+				.Logging,
 			],
 			plugins: [
 				.plugin(name: "SwiftLint", package: "SwiftLintPlugin"),
@@ -128,6 +128,7 @@ let package = Package(
 				.auth,
 				.eventProducer,
 				.GRDB,
+				.AnyCodable,
 			],
 			resources: [
 				.process("README.md"),
--- a/Package.swift
+++ b/Package.swift
@@ -72,6 +72,7 @@ let package = Package(
 			name: "Common",
 			dependencies: [
 				.Logging,
+				.AnyCodable,
 			],
 			plugins: [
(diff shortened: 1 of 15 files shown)


asendra:
I need to merge another PR today in a bit. Would you like for me to make a new release with both things for today's cutoff?

If you don't want it for today, don't merge it until tomorrow 😀 !

andrelouw:
@asendra would be awesome if you can include this in the new version 💪

PR #110, opened 2024-10-14
add FF for not performing any action when item reaches end

ceciliasaraiva:
I'm investigating different issues we have in Player. One of them seems to be related to our setup where we add `player.actionAtItemEnd = .advance`. By changing this, I was not able to reproduce the bug with seek. So I'd like to try out how it behaves when we set this as `player.actionAtItemEnd = .none`. So added another FF so we can test this.

I continuing testing this to see if I'm missing some specific scenarios.


asendra (approved):
Nice find!
We can (should!) test this on production, but before we go and enable it for all and remove the flag we need to make how gapless playback is affected (if it is)

ceciliasaraiva:
> Nice find! We can (should!) test this on production, but before we go and enable it for all and remove the flag we need to make how gapless playback is affected (if it is)

I'm testing and it seems fine but QA will probably do a better job than me.

PR #111, opened 2024-10-14
Logging: make source optional + added documentation

e-kononenko:
Make source optional + added documentation for `TidalLoggable` and for the logging function


Sources/Common/Logger/TidalLogger.swift
```diff
@@ -29,9 +29,21 @@ public extension TidalLogger {
...
+	///   - message: Message you want to log
+	///   - source: The origin of a log message. If you don't provide it, it will be set automatically to the name of a calling module, which might not be intended
+	///   - level: Level of a log message
+	///   - metadata: Additional information attached to a log message
+	///   - file: The file this log message originates from (there's usually no need to pass it explicitly as it
+	///            defaults to `#fileID`.
+	///   - function: The function this log message originates from (there's usually no need to pass it explicitly as
+	///                it defaults to `#function`).
+	///   - line: The line this log message originates from (there's usually no need to pass it explicitly as it
+	///            defaults to `#line`).
```
ceciliasaraiva, line 43: [suggestion] we can probably get rid of the space
```suggestion
	/// defaults to `#fileID`.
	///   - function: The function this log message originates from (there's usually no need to pass it explicitly as
	/// it defaults to `#function`).
	///   - line: The line this log message originates from (there's usually no need to pass it explicitly as it
	/// defaults to `#line`).
```

ceciliasaraiva (approved):
just a minor suggestion for the documentation 🤓

PR #112, opened 2024-10-15
[PLAYER] Reworked AVQueuePlayerWrapper to solve enqueue crash

asendra:
## Context
Adjusted the implementation of `AVQueuePlayerWrapper` that uses the improved caching system to solve a crash we have seen on production when we enable the flag to a % of our users.

Unfortunately, we have been unable to reproduce this crash locally.

We have been testing for the past few days a build using this PR branch with ~20 users and ~4k streams without the crash happening and without other issues appearing.

## Description

This PR contains two main changes to the Player <-> cache integration
* The cache uses it's own queue
* We enqueue immediately the first item we load even if it's not cached, instead of waiting until the play is called (this was done to have more time to start caching it)

### DIFF
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
@@ -2,9 +2,9 @@ import AVFoundation
 import Foundation
 import MediaPlayer
 
-// swiftlint:disable file_length
-
 // MARK: - Constants
+
+// swiftlint:disable file_length
 
 private enum Constants {
 	static let defaultVolume: Float = 1
@@ -62,7 +62,7 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
 		queue.maxConcurrentOperationCount = 1
 		queue.qualityOfService = .userInitiated
 
-		assetFactory = AVURLAssetFactory(with: queue)
+		assetFactory = AVURLAssetFactory()
 		player = AVQueuePlayerWrapper.createPlayer()
 
 		preparePlayer()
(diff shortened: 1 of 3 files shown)


Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
```diff
@@ -594,19 +590,17 @@ private extension AVQueuePlayerWrapper {
...
 			if let currentPlayerItem = self.player.currentItem {
 				self.delegates.completed(asset: asset)
-
 				if self.player.timeControlStatus == .playing {
 					self.playing(playerItem: currentPlayerItem)
 				}
-
 			} else {
-				// AVPlayer had no other item in the queue (we are still downloading the next one)
+				// AVPlayer had no other item in the queue (we are still downloading the next one?)
```
ziad-halabi9, line 597: Is this comment a question now?

asendra: he, unclear what I mean there lol.

There is an `if/else` right after. So it depends on the `if` condition.

Would be better like this I guess
```
// AVPlayer had no other item in the queue (We might still be downloading the next one)`
```

asendra: The reason for this is that if an `AVPlayerItem` is in the player queue, it will not be cached in advance. I'm not sure why, but these are both internal Apple APIs...

So what I do internally when `setNext` is called, is to start caching it when the current item has fully downloaded), but not enqueue it right away.

ziad-halabi9 (approved):

ziad-halabi9 (approved):

PR #114, opened 2024-10-16
[Player] Fix OfflineItem size calculation

asendra:
I noticed a couple of bugs in the size calculation code while testing the client integration.

This PR fixes them, and refactors the size calculation into the `DownloadTask` , where it makes more sense.

The two things fixed:
* Use the correct size algorithm for the calculation based on the download type (using PlaybackInfo), 
* Use the final URL where we move the progressive file download into, instead of the temporary download path! 🤦‍♂️

### DIFF
--- a/Sources/Player/OfflineEngine/Internal/DownloadTask.swift
+++ b/Sources/Player/OfflineEngine/Internal/DownloadTask.swift
@@ -93,11 +93,7 @@ final class DownloadTask {
 
 	func setMediaUrl(_ url: URL) {
 		localAssetUrl = url
-		finalize()
-	}
-
-	func setSize(_ size: Int) {
-		localAssetSize = size
+		localAssetSize = calculateSize()
 		finalize()
 	}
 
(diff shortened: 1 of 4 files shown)


ziad-halabi9 (approved):

Sources/Player/OfflineEngine/Internal/DownloadTask.swift
```diff
@@ -127,6 +123,54 @@ final class DownloadTask {
...
+		) {
+			for case let fileURL as URL in enumerator {
+				do {
+					let fileAttributes = try fileURL.resourceValues(forKeys: [URLResourceKey.fileSizeKey])
+					if let fileSize = fileAttributes.fileSize {
+						totalSize += Int(fileSize)
+					}
+				} catch {
+					PlayerWorld.logger?.log(loggable: PlayerLoggable.failedToCalculateSizeForHLSDownload(error: error))
+				}
```
ceciliasaraiva, line 155: [suggestion] Minor, I think if we use `reduce` we would reduce the indentation by 1, readability would be mostly same I guess.

ceciliasaraiva (approved):
nice, just added a minor comment.

Sources/Player/OfflineEngine/Internal/DownloadTask.swift
asendra: valid suggestion. I have a few stacked PRs so I think I won't bother as the difference is minimal. 
But I appreciate it!

PR #116, opened 2024-10-16
Add `Common` product

mercedesbunz:
To enable using the contents of `Common` individually it needs to be exposed as a product

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -30,6 +30,10 @@ let package = Package(
 			name: "Player",
 			targets: ["Player"]
 		),
+		.library(
+			name: "Common",
+			targets: ["Common"]
+		),
 	],
 	dependencies: [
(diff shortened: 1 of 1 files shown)


e-kononenko (approved):
Thank you! Somehow it worked partially before and that's why we didn't notice it, but with this change it should be more stable!

asendra:
cc @mercedesbunz @e-kononenko 

Sorry, I got to this too late it seems 🫤

I don't think it was not working. Previously we had decided that `Common` would not be needed alone by external consumers of the SDK.
So, to prevent them from needing to do
```
import Common
import Player
```

We were exposing `Common` implicitly with every module by doing this 
in `PlayerExports.swift`
```
@_exported import Common
```

Allowing them to just do 
```
import Player
```


Has anything changed recently that causes external consumers of the SDK to only need to import `Common`?

PR #117, opened 2024-10-17
[Player] Refactor OfflineEngine Listener setup and handling

asendra:
Refactored and renamed `OfflinerDelegate` into `OfflineEngineListener`, and added it to the bootstrap process.
Now it also ensures it will use the same queue for the notification events as the `PlayerListener`

One possibility to discuss would be to have different queues for each listener and instead of using `NotificationsHandler` for both, have a different class for each.

### DIFF
--- a/Sources/Player/Common/Event/NotificationsHandler.swift
+++ b/Sources/Player/Common/Event/NotificationsHandler.swift
@@ -2,10 +2,12 @@ import Foundation
 
 final class NotificationsHandler {
 	private let listener: PlayerListener
+	private let offlineEngineListener: OfflineEngineListener?
 	private let queue: DispatchQueue
 
-	init(listener: PlayerListener, queue: DispatchQueue) {
+	init(listener: PlayerListener, offlineEngineListener: OfflineEngineListener, queue: DispatchQueue) {
 		self.listener = listener
+		self.offlineEngineListener = offlineEngineListener
 		self.queue = queue
 	}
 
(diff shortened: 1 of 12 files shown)


Sources/Player/Common/Event/NotificationsHandler.swift
```diff
@@ -58,4 +60,40 @@ final class NotificationsHandler {
...
 		}
 	}
+
+	func offliningStarted(for mediaProduct: MediaProduct) {
+		queue.async {
+			self.offlineEngineListener?.offliningStarted(for: mediaProduct)
+		}
+	}
+
+	func offliningProgress(for mediaProduct: MediaProduct, is downloadPercentage: Double) {
```
ceciliasaraiva, line 70: [suggestion] maybe for consistency with the other names where we use "started", "completed", etc.
```suggestion
	func offliningProgressed(for mediaProduct: MediaProduct, is downloadPercentage: Double) {
```

Sources/Player/Common/Event/NotificationsHandler.swift
```diff
@@ -2,10 +2,12 @@ import Foundation
 
 final class NotificationsHandler {
 	private let listener: PlayerListener
```
ceciliasaraiva, line 4: [question] WDYT about renaming this to `playerListener`, since now we have more than one "listener" in the class.

Sources/Player/Common/Event/NotificationsHandler.swift
```diff
@@ -58,4 +60,40 @@ final class NotificationsHandler {
 			self.listener.djSessionTransitioned(to: transition)
 		}
 	}
+
+	func offliningStarted(for mediaProduct: MediaProduct) {
+		queue.async {
```
ceciliasaraiva, line 65: [question] since this is "new", have you considered using structured concurrency? So it's potentially less code to migrate when (one sunny day) we eventually move to it 🤓 or maybe a separate queue would help already, so we could migrate separately if possible.

ceciliasaraiva (approved):
added few comments

Sources/Player/Common/Event/NotificationsHandler.swift
asendra: Great suggestion

Sources/Player/Common/Event/NotificationsHandler.swift
asendra: Yeah, I had though about that! 
I will do it.

PR #120, opened 2024-10-18
[Player] Rename HLS assets in OfflineEngine

asendra:
Change the naming of the downloaded HLS AVURLAssets to match the naming used in the cache.
This naming will better reflect what is actually stored in case a user navigates to the Storage settings on their device.

At some later point, we might decide to bridge the user-facing name of the MediaProduct (and the artwork!), which we could use here.

## Screenshots 📷 
IMG_3092A5C89072-1


ceciliasaraiva (approved):
nice

PR #122, opened 2024-10-18
[Player] Make OfflineEngine non-optional

asendra:
Now that we are getting ready to test the new Offline Engine, we need to have it always initialized because there can be cases where someone downloads stuff with it, but then it gets the FF flag disabled for it, and it needs to still be able to play them.

### DIFF
--- a/Sources/Player/Player.swift
+++ b/Sources/Player/Player.swift
@@ -21,10 +21,10 @@ public final class Player {
 	private(set) var playerEngine: PlayerEngine
 
 	@Atomic
-	private(set) var offlineStorage: OfflineStorage?
+	private(set) var offlineStorage: OfflineStorage
 
 	@Atomic
-	private(set) var offlineEngine: OfflineEngine?
+	private(set) var offlineEngine: OfflineEngine
 
 	// MARK: - Properties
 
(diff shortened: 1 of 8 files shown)


Tests/PlayerTests/Player/PlayerTests.swift
```diff
@@ -50,6 +50,11 @@ final class PlayerTests: XCTestCase {
 		)
 
 		let notificationsHandler = NotificationsHandler.mock()
+		let offlineEngine = OfflineEngine.mock(
+			storage: storage,
```
ceciliasaraiva, line 54: [question] should we have an offline mocked storage? I see we are using `GRDBOfflineStorage`

Sources/Player/Player.swift
```diff
@@ -122,6 +122,10 @@ public extension Player {
 			return nil
 		}
 
+		guard let offlineStorage = Player.initializedOfflineStorage() else {
```
ceciliasaraiva, line 125: [question] we already have log for this one, right?

ceciliasaraiva (approved):
just two questions

Tests/PlayerTests/Player/PlayerTests.swift
asendra: We could mock the storage directly, but we are already mocking the DB that `GRDBOfflineStorage` uses so it is in memory.

So it should not be flaky and slow. 
I would only invest the time to add an `OfflineStorage` mock when we add or modify tests that need it.

Sources/Player/Player.swift
asendra: Yes, internally we do a `PlayerLoggable.withDefaultDatabase(error: error)` if we fail to initialize the Database.

PR #123, opened 2024-10-18
Add TidalAPI module

e-kononenko:
New module with the generated API


Package.swift
```diff
@@ -66,6 +70,15 @@ let package = Package(
 				.plugin(name: "SwiftLint", package: "SwiftLintPlugin"),
 			]
 		),
+		.target(
+			name: "API",
+			dependencies: [
+				.AnyCodable,
```
mercedesbunz, line 76: I think this also needs `Auth` as a dependency

Package.swift
```diff
@@ -34,6 +34,10 @@ let package = Package(
 			name: "Common",
 			targets: ["Common"]
 		),
+		.library(
+			name: "API",
```
mercedesbunz, line 38: Maybe we want to have a more descriptive name here since when importing it is just
```swift
import API
```

WDYT?

michpohl: How about we call it `TidalApi`? That seems to be the easiest way to reduce confusion...

Sources/API/Utils/APIError.swift
michpohl, line 1: I wonder if we should make this extend our already existing `TidalError` for better compatibility?

Sources/API/Utils/OpenAPIClientAPI+Auth.swift
```diff
@@ -0,0 +1,5 @@
+import Auth
+
+extension OpenAPIClientAPI {
```
michpohl, line 3: This name is kinda odd 😄

Sources/API/generate_and_clean.sh
michpohl, line 1: I did what modern devs do and asked ChatGPT what it thinks about the script 😄 
Here is the answer. Do with it what you fell like. It's fine by me do ignore it.
```
#!/bin/bash

# Define the input and output directories
input_dir="Config/input"
output_dir="Generated"

# Step 1: Clear the contents of the input directory
rm -rf "$input_dir"/*
mkdir -p "$input_dir"

# Step 2: Download the JSON files and save them in the input directory
echo "Downloading API specs from developer.tidal.com"
curl -f -o "$input_dir/tidal-catalog-v2-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-catalogue-jsonapi/tidal-catalog-v2-openapi-3.0.json || { echo "Failed to download tidal-catalog-v2-openapi-3.0.json"; exit 1; }
curl -f -o "$input_dir/tidal-search-v2-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-search-jsonapi/tidal-search-v2-openapi-3.0.json || { echo "Failed to download tidal-search-v2-openapi-3.0.json"; exit 1; }
curl -f -o "$input_dir/tidal-user-v2-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-user-jsonapi/tidal-user-v2-openapi-3.0.json || { echo "Failed to download tidal-user-v2-openapi-3.0.json"; exit 1; }
curl -f -o "$input_dir/tidal-user-content-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-user-content/tidal-user-content-openapi-3.0.json || { echo "Failed to download tidal-user-content-openapi-3.0.json"; exit 1; }

# Step 3: Clear the contents of the output directory
echo "Clearing the contents of the output directory"
rm -rf "$output_dir"/*
mkdir -p "$output_dir"

# Step 4: Loop through all JSON files in the input folder
for json_file in "$input_dir"/*.json; do
  # Create a unique output directory for each file
  file_output_dir="$output_dir/$(basename "$json_file" .json)"
  mkdir -p "$file_output_dir"

  # Run the OpenAPI generator for each JSON file
  openapi-generator generate \
    -i "$json_file" \
    -g swift5 \
    -o "$file_output_dir" \
    -c Config/openapi-config.yml || { echo "OpenAPI generation failed for $json_file"; exit 1; }

  # Step 5: Remove the unwanted files
  rm -f "$file_output_dir"/{Package.swift,Cartfile,git_push.sh,project.yml,OpenAPIClient.podspec,.swiftformat,.gitignore,.openapi-generator,.openapi-generator-ignore}

  echo "Generation and cleanup completed successfully for $(basename "$json_file")."
done
```

> Changes:

	1.	Error Handling: curl -f and error checks were added to gracefully handle download and OpenAPI generation failures.
	2.	Parallelization: You could run the generation in parallel if you choose to, though it is not applied in this version.
	3.	Unique Output Directories: Separate output directories are created for each JSON file to prevent overwriting files.
	4.	Cleaner Cleanup: Used a single rm command to remove unwanted files.

Sources/API/Generated/.openapi-generator/FILES
```diff
@@ -0,0 +1,198 @@
+.gitignore
+.openapi-generator-ignore
+.swiftformat
+Cartfile
+OpenAPIClient.podspec
+OpenAPIClient/Classes/OpenAPIs/APIHelper.swift
```
michpohl, line 6: the plural `OpenAPIs` looks odd... 🤷 
I could also see it making sense to put the generated files into a `Generated` folder, so it's easy to see what's handmade and what not?

michpohl (commented):
I added a few comments, from a high level perspective. I can't properly judge the swift implementation details, but overall this looks good as a starting point. 👍

Sources/API/generate_and_clean.sh
mercedesbunz: Its funny because the script was created by ChatGPT in the first place 😆  But I see this behaviour a lot like when you ask it: "What could be optimised?" or whatever it will always find something but step 3. is actually what we want to NOT have it do it. 

But since step 4 is called `Cleaner cleanup` we for sure need it 😆 Although I think it gets harder to read

Sources/API/Generated/.openapi-generator/FILES
mercedesbunz: It actually will write all the generated code into `Generated` as you can see in the path of this file but you are right and it seems it does not pick that up for the creation of this `FILES` file

Package.swift
e-kononenko: I had the same thought 🙂 

Thank you, will do!

Sources/API/generate_and_clean.sh
e-kononenko: This is the thing that we discussed: to use the same folder and let generated models to be overwritten, in order to avoid collisions. Maybe we could add a comment about this to the script? 🤔 

Obviously, because of this, the generation runs should be consecutive 🙂

Sources/API/Generated/OpenAPIClient/Classes/OpenAPIs/APIs/AlbumsAPITidal.swift
```diff
@@ -0,0 +1,87 @@
...
+#endif
+
+/// This is a wrapper around `AlbumsAPI` that uses the injected credentialsprovider
+/// from `OpenAPIClientAPI.credentialsProvider` to provide a convenience API.
+///
+/// Usage example:
+/// ```swift
+/// OpenAPIClientAPI.credentialsProvider = TidalAuth.shared
+/// let dataDocument = try await AlbumsAPI.getResource()
+/// ```
```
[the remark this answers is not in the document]
e-kononenko: oh yes, we should adjust the template to have:

 `/// let dataDocument = try await AlbumsAPITidal.getResource()`

Sources/API/Utils/OpenAPIClientAPI+Auth.swift
e-kononenko: Agree, I will try to find another name 🙂

Package.swift
e-kononenko: I wouldn't camel-case the "API" part though 🙂

Package.swift
e-kononenko: oh sure, thank you!

Sources/API/generate_and_clean.sh
michpohl: By the way, for Android, I had to do a different approach: Merge the json files into one and then do the generation, because I need the generator to have full context when building the resource types.

What I do basically is join the schema files and throw out any duplicates (as we established, we can expect them to be the same). Then after, I only need to do a single generator run. 

This could be adopted here too, if you're interested. It isn't needed for you, but just so you know what's available 😄

Sources/API/Generated/.openapi-generator/FILES
michpohl: ah yes you're right! I don't care too much about this file tbh. What do you guys think?

Package.swift
michpohl: You wouldn't, I would 🤣 
image

mercedesbunz (approved):
Checked with the integration branch and saw it working 👏

Sources/API/Utils/OpenAPIClientAPI+Auth.swift
e-kononenko: I decided to do it later - because it is a default behaviour and we need to have one more overwritten template file. It requires to make a lot of changes in the integration as well, so the scope of the task will be significantly increased.

Sources/API/Generated/.openapi-generator/FILES
e-kononenko: honestly, I didn't even notice this `.openapi-generator` folder and its content 😅 

I don't really think they are needed, for me it looks like a kind of meta info, which is generated every time and everything would work without it... so maybe we can just remove it?

Sources/API/Utils/APIError.swift
e-kononenko: Agree, but we can do it later. I will create a ticket with all fine-tunings discussed in this PR and we can do them later. At the moment we have MVP which works fine with our integration 🙂

Package.swift
e-kononenko: Thank you! I will definitely follow this advice when I have a chance to commit to Java/Kotlin repos 😄 

https://www.swift.org/documentation/api-design-guidelines/#general-conventions 

Screenshot 2024-10-22 at 16 24 38

Sources/TidalAPI/Utils/APIError.swift
```diff
@@ -0,0 +1,61 @@
+import Foundation
+
+public struct APIError: LocalizedError, Equatable {
```
asendra, line 3: Now that you renamed the package from `API` to `TidalAPI`, this should be `TidalAPIError`?

Sources/TidalAPI/generate_and_clean.sh
```diff
@@ -0,0 +1,44 @@
...
+echo "Dowloading API specs from developer.tidal.com"
+curl -o "$input_dir/tidal-catalog-v2-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-catalogue-jsonapi/tidal-catalog-v2-openapi-3.0.json
+curl -o "$input_dir/tidal-search-v2-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-search-jsonapi/tidal-search-v2-openapi-3.0.json
+curl -o "$input_dir/tidal-user-v2-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-user-jsonapi/tidal-user-v2-openapi-3.0.json
+curl -o "$input_dir/tidal-user-content-openapi-3.0.json" https://developer.tidal.com/apiref/api-specifications/api-public-user-content/tidal-user-content-openapi-3.0.json
+
+# Step 3: Clear the contents of the output directory
+echo "Clear the contents of the output directory"
+rm -rf "$output_dir"/*
+mkdir -p "$output_dir"
```
asendra, line 21: Maybe this has been talked about already.

If we clean up the folder, and the new generated code has the same things as before, plus just one new endpoint (for example) would git detect everything as new, or would it detect just the new endpoint change? 

Would be great to have cleaner commits.

asendra (approved):
Left some minor questions, but I don't have much input here. The script and the structure looks good to me.

Sources/TidalAPI/Utils/APIError.swift
e-kononenko: Good point!

Sources/TidalAPI/generate_and_clean.sh
e-kononenko: > Would be great to have cleaner commits

well, we did most of the work in another (private) repo, so commits here can be too generic, I suppose 🙂 

Answering your question: it shows only changes.

asendra: no no, I was not talking about the commits on this PR! 😁 

I was referring to the future commits when the generator updates things. But if it shows only changes, that is perfect! 👏

michpohl (approved):

PR #127, opened 2024-10-22
[Player] Fix playback of certain offlined items

asendra:
In some cases, the client will request to play a productId, but the backend will return an alternative one depending on the user configuration. 

The most simple example is when trying to play/download an Atmos track, with the Dolby Atmos setting off. this will cause the backend to return an alternative `STEREO` version with a different track Id.

This broke offline playback for those items because we were not finding it in the offline storage.


ceciliasaraiva (approved):

Sources/Player/OfflineEngine/Internal/Storage/OfflineEntry.swift
```diff
@@ -92,13 +95,15 @@ struct OfflineEntry: Codable {
 	}
 
 	init(
+		for mediaProduct: MediaProduct,
```
ziad-halabi9, line 98: Small suggestion, if the productId is all that is needed from the MediaProduct, then it's better not to rely on the full Object as an argument but then just the id.

asendra: Fair! 👍 
I wanted to make it clear that the new `productId` was coming from `MediaProduct`, but it is not a good practice to pass all that extra context around.

ziad-halabi9 (approved):

Sources/Player/OfflineEngine/Internal/Storage/OfflineEntry.swift
asendra: Made the change!

PR #129, opened 2024-10-23
[Player] Fix AVQueuePlayerWrapper.canPlay() for PlayableOfflinedMediaProduct 

asendra:
Although the functionality was working correctly, it was by luck. We had a sily mistake where we weren't properly identifying `PlayableOfflinedMediaProduct` plays comming from our OfflineEngine as `offlined`, so the check we were doing inside `AVQueuePlayerWrapper.canPlay()` was incorrect once that was fixed.

This PR solves both issues:
* Correctly identify `PlayableOfflinedMediaProduct` plays as `isOfflined: true`
* Change the check for `FLAC` files to use the `MediaType` of the file to decide if it can be played or not

### DIFF
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
@@ -88,7 +88,9 @@ final class AVQueuePlayerWrapperLegacy: GenericMediaPlayer {
 			guard let codec else {
 				return false
 			}
-			return supportedCodecs.contains(codec) || (codec == .FLAC && !isOfflined)
+			// Because we need to support items downloaded with both the legacy and the new OfflineEngine
+			// We can't just add .FLAC to the 'supportedCodecs' array, and we need to depend on the mediaType
+			return supportedCodecs.contains(codec) || (codec == .FLAC && mediaType != MediaTypes.BTS)
 		}
 	}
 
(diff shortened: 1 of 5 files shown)


Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
```diff
@@ -88,7 +88,9 @@ final class AVQueuePlayerWrapperLegacy: GenericMediaPlayer {
 			guard let codec else {
 				return false
 			}
-			return supportedCodecs.contains(codec) || (codec == .FLAC && !isOfflined)
+			// Because we need to support items downloaded with both the legacy and the new OfflineEngine
+			// We can't just add .FLAC to the 'supportedCodecs' array, and we need to depend on the mediaType
+			return supportedCodecs.contains(codec) || (codec == .FLAC && mediaType != MediaTypes.BTS)
```
e-kononenko, line 93: does it make sense to extract this functionality from here and from `AVQueuePlayerWrapper` and avoid this logic duplication?

e-kononenko (approved):

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift
asendra: `AVQueuePlayerWrapperLegacy` will be removed in a few weeks if the cache fixes have worked 🤞 . 
A feature flag right now decides whether we use one or the other.

PR #133, opened 2024-10-27
Add player item monitor for when an item reaches the end when action at end is none

ceciliasaraiva:
Continuation of logic for when the item reaches the end.
First part here: https://github.com/tidal-music/tidal-sdk-ios/pull/110

Here, I've added a monitor for when the player item reached the end. At that moment, since we don't automatically perform any action automatically, I remove it from the player when the FF is enabled.

I continue testing to see if there's any scenario which I might have missed.

---

This pull request introduces a new feature to handle the event when an `AVPlayerItem` finishes playing. It adds a new monitor class and integrates it into the existing player wrapper classes.

### New Feature: Item Played to End Monitoring

* **New Monitor Class**:
  * [`Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Observers/ItemPlayedToEndMonitor.swift`](diffhunk://#diff-66dd1acaa94ad9d8f2e1e105b63057e7e306dcb88df07d44e47a84f294b3e813R1-R35): Introduced `ItemPlayedToEndMonitor` to observe when an `AVPlayerItem` finishes playing and trigger a callback.

* **Integration into `AVPlayerItemMonitor`**:
  * [`Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVPlayerItemMonitor.swift`](diffhunk://#diff-2685a78fb53c72072cf13c7cb75694b64150414a78498eaac68c4fc20c736327R11-R27): Added `playerItemDidPlayToEndTimeMonitor` to monitor when an item has played to the end.

### Updates to Player Wrappers

* **`AVQueuePlayerWrapper`**:
  * [`Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift`](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R404): Updated to handle the `onItemPlayedToEnd` callback and remove the player item if a feature flag is set. [[1]](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R404) [[2]](diffhunk://#diff-b75f24d9a1db1bbe79fed5b71b9fd741a62df41c29d49643af0db04b36fe6985R635-R640)

* **`AVQueuePlayerWrapperLegacy`**:
  * [`Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Legacy/AVQueuePlayerWrapperLegacy.swift`](diffhunk://#diff-f2fb496a12a5d9bea6f0efb2ea3766e9213555c80f56ad9f7c7dc4cc7d255917R375): Similarly updated to handle the `onItemPlayedToEnd` callback and remove the player item if a feature flag is set. [[1]](diffhunk://#diff-f2fb496a12a5d9bea6f0efb2ea3766e9213555c80f56ad9f7c7dc4cc7d255917R375) [[2]](diffhunk://#diff-f2fb496a12a5d9bea6f0efb2ea3766e9213555c80f56ad9f7c7dc4cc7d255917R534-R535) [[3]](diffhunk://#diff-f2fb496a12a5d9bea6f0efb2ea3766e9213555c80f56ad9f7c7dc4cc7d255917R566-R571)

### DIFF
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Observers/ItemPlayedToEndMonitor.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Observers/ItemPlayedToEndMonitor.swift
@@ -1,0 +1,35 @@
+import AVFoundation
+import Foundation
+
+final class ItemPlayedToEndMonitor {
+	private let playerItem: AVPlayerItem
+	private let onPlayedToEnd: (AVPlayerItem) -> Void
+
+	init(playerItem: AVPlayerItem, onPlayedToEnd: @escaping (AVPlayerItem) -> Void) {
+		self.playerItem = playerItem
+		self.onPlayedToEnd = onPlayedToEnd
+
+		NotificationCenter.default.addObserver(
+			self,
+			selector: #selector(playerItemDidPlayToEndTime(notification:)),
+			name: NSNotification.Name.AVPlayerItemDidPlayToEndTime,
+			object: playerItem
+		)
+	}
+
+	deinit {
+		NotificationCenter.default.removeObserver(
+			self,
+			name: NSNotification.Name.AVPlayerItemDidPlayToEndTime,
+			object: playerItem
+		)
+	}
+
+	@objc
+	private func playerItemDidPlayToEndTime(notification: Notification) {
+		if let playerItem = notification.object as? AVPlayerItem {
+			print("--> PlayerItemDidPlayToEndTimeMonitor.playerItemDidPlayToEndTime(notification: \(notification))")
+			onPlayedToEnd(playerItem)
+		}
+	}
+}
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVPlayerItemMonitor.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVPlayerItemMonitor.swift
@@ -8,6 +8,7 @@ final class AVPlayerItemMonitor {
 	private let completelyDownloadedMonitor: CompletelyDownloadedMonitor
 	private let readyToPlayMonitor: ReadyToPlayMonitor
 	private let inStreamMetadataMonitor: DJSessionTransitionMonitor
+	private let playerItemDidPlayToEndTimeMonitor: ItemPlayedToEndMonitor
 
 	init(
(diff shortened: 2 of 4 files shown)


Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Observers/ItemPlayedToEndMonitor.swift
```diff
@@ -0,0 +1,35 @@
...
+			self,
+			name: NSNotification.Name.AVPlayerItemDidPlayToEndTime,
+			object: playerItem
+		)
+	}
+
+	@objc
+	private func playerItemDidPlayToEndTime(notification: Notification) {
+		if let playerItem = notification.object as? AVPlayerItem {
+			print("--> PlayerItemDidPlayToEndTimeMonitor.playerItemDidPlayToEndTime(notification: \(notification))")
```
asendra, line 31: Should this be deleted, or convert into a proper Player Log?

asendra (approved):
With this new event and flag, does `playerItemChanged(oldPlayerItem: AVPlayerItem)` get called?

The way I understand this works is that the current item ends, it is removed and we call `completed()` to the delegates, which ends up in `PlayerEngine.completed() `calling play in the next(new current) item.
Is that correct?

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/Observers/ItemPlayedToEndMonitor.swift
ceciliasaraiva: no, deleted, I always add a lot of prints to get the whole flow, will delete it.

ceciliasaraiva:
> With this new event and flag, does `playerItemChanged(oldPlayerItem: AVPlayerItem)` get called?
> 
> The way I understand this works is that the current item ends, it is removed and we call `completed()` to the delegates, which ends up in `PlayerEngine.completed() `calling play in the next(new current) item. Is that correct?

It's still being called. When we remove the old item from player, it's same as advancing to the next item, my guess is that when we call remove again in the `playerItemChanged`, if not present, nothing happens and then it proceeds to completed as you described above.

PR #141, opened 2024-10-28
[Player] Improve external player injection

asendra:
Their PR changes how external players are provided so they can be changed at runtime.

Currently, we were passing an array at `bootstrap` that can't be updated, and this caused issues with our improved caching feature flag.


ceciliasaraiva (approved):
nice!

PR #170, opened 2024-11-20
Fixed Auth Login Flow in AuthTestApp

asendra:
Seems like recent changes in how `SFSafariViewController` handles redirects broke the sample OAuth login flow in the `AuthTestApp`

This PR updates the flow to use an `ASWebAuthenticationSession` and supports both custom URL schemes for the redirection or a regular HTTPS URL if it has been configured with the associated domains feature.


deaa (approved):

TestApps/Auth/AuthTestApp/AuthViewModel.swift
```diff
@@ -102,7 +118,55 @@ class AuthViewModel: ObservableObject {
...
+			expiresIn = duration.formatted(.time(pattern: .minuteSecond(padMinuteToLength: 2)))
+		} else {
+			let timeInterval = expiresAt.timeIntervalSinceNow
+			let minutes = Int(timeInterval) / 60
+			let seconds = Int(timeInterval) % 60
+			expiresIn = String(format: "%02d:%02d", minutes, seconds)
+		}
+	}
+
+	func presentationAnchor(for session: ASWebAuthenticationSession) -> ASPresentationAnchor {
```
ceciliasaraiva, line 132: [Question] maybe we need to provide another class which is not a `MainActor`?

asendra: Yes, that works! thanks 👍

PR #171, opened 2024-11-26
[Player] Fix initialization errors due to file Protection in OfflineStorage DB

asendra:
## Context
We noticed some crashes in our TIDAL client happening under some circumstances, which were caused due to an uninitialized Player instance.

After some investigation, the reason why the Player was not being initialised properly was due to an error opening the `OfflineStorage` database.

We are already logging when this happens, unfortunately, this was happening before we had initialized the `PlayerLogger` 😅 

If we had the logs in place, we probably would see that the issue is due to the db file having `fileProtection = .completeFileProtectionUntilFirstUserAuthentication`. When the app gets prewarmed after a reboot, but before the user has unlocked his device for the first time, this files are not accesible, and `Player.bootstrap()` will fail, causing a crash in our client because it expects it to never be `nil`

## Fix

This PR changes three main things
- Init the PlayerLogger first thing in the bootstrap process, 
- Clarifying the error handling for `OfflineStorage`
- Setting the file protection to `.none` for the `OfflineStorage` database

### DIFF
--- a/Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
+++ b/Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
@@ -40,17 +40,12 @@ class GRDBOfflineStorage {
 		}
 	}
 
-	static func withDefaultDatabase() -> GRDBOfflineStorage? {
-		do {
-			let databaseURL = try GRDBOfflineStorage.databaseURL()
-			let dbQueue = try DatabaseQueue(path: databaseURL.path)
-			try GRDBOfflineStorage.initializeDatabase(dbQueue: dbQueue)
+	static func withDefaultDatabase() throws -> GRDBOfflineStorage {
+		let databaseURL = try GRDBOfflineStorage.databaseURL()
+		let dbQueue = try DatabaseQueue(path: databaseURL.path)
+		try GRDBOfflineStorage.initializeDatabase(dbQueue: dbQueue)
 
-			return GRDBOfflineStorage(dbQueue: dbQueue)
-		} catch {
-			PlayerWorld.logger?.log(loggable: PlayerLoggable.withDefaultDatabase(error: error))
-			return nil
-		}
+		return GRDBOfflineStorage(dbQueue: dbQueue)
 	}
 }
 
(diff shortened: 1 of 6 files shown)


Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
```diff
@@ -122,7 +117,22 @@ private extension GRDBOfflineStorage {
 	static func databaseURL() throws -> URL {
 		let appSupportURL = PlayerWorld.fileManagerClient.applicationSupportDirectory()
 		let directoryURL = appSupportURL.appendingPathComponent("PlayerOfflineDatabase", isDirectory: true)
-		try PlayerWorld.fileManagerClient.createDirectory(at: directoryURL, withIntermediateDirectories: true)
+		if PlayerWorld.fileManagerClient.fileExists(atPath: directoryURL.path, isDirectory: nil) {
+			var attributes = try? PlayerWorld.fileManagerClient.attributesOfItem(directoryURL.path)
+			if attributes != nil,
```
e-kononenko, line 122: `if let attributes` maybe? Then you don't need to force unwrap them later

Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
```diff
@@ -122,7 +117,22 @@ private extension GRDBOfflineStorage {
 	static func databaseURL() throws -> URL {
 		let appSupportURL = PlayerWorld.fileManagerClient.applicationSupportDirectory()
 		let directoryURL = appSupportURL.appendingPathComponent("PlayerOfflineDatabase", isDirectory: true)
-		try PlayerWorld.fileManagerClient.createDirectory(at: directoryURL, withIntermediateDirectories: true)
+		if PlayerWorld.fileManagerClient.fileExists(atPath: directoryURL.path, isDirectory: nil) {
+			var attributes = try? PlayerWorld.fileManagerClient.attributesOfItem(directoryURL.path)
```
e-kononenko, line 121: shouldn't we also log an error if this call throws it?

Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
```diff
@@ -122,7 +117,22 @@ private extension GRDBOfflineStorage {
...
+			if attributes != nil,
+			   let protectionKey = attributes?[FileAttributeKey.protectionKey] as? FileProtectionType,
+			   protectionKey != FileProtectionType.none
+			{
+				attributes?[FileAttributeKey.protectionKey] = FileProtectionType.none
+				try? PlayerWorld.fileManagerClient.setAttributes(attributes!, directoryURL.path)
```
e-kononenko, line 127: maybe log this as well if throws?

Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
asendra: I need to modify it, so it needs to be a `var` 🤔 

I tried a few different ways, and I did not like any of them...

Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
asendra: yes, we can log it. 
I just don't want it to cause a crash. If it fails, execution should continue normally.

Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
asendra: yeah. I might combine both in a single log. No need to differentiate really (especially if I attach the error description).

Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/GRDBOfflineStorage.swift
asendra: If I add the log, I can do `try` instead of `try?`, and it improves the readability of the code a lot.

asendra: done!

e-kononenko (approved):

PR #173, opened 2024-11-27
[Player] Fix output sample rate (mainly for DACs)

asendra:
## Context
With the new `Player` implementations used after the Caching v2 project, we disabled behind a development flag the feature of reading the asset metadata.

Unfortunately, we forgot to report the asset metadata provided by the backend instead, which resulted in the `AVAudioSession` output sample rate not being updated as the user switches tracks.

## Description
Report output sample rate from BE even if we are not verifying it from the asset


ceciliasaraiva (approved):
nice and sorry for the formatting fixes, seems like I continue to have issues in this repo 🤔

PR #174, opened 2024-11-28
[Player] Remove AVQueuePlayerWrapperLegacy implementation

asendra:
Now that we are exclusively using the new `AVQueuePlayerWrapper` implementation with the improved caching, we can get rid of the legacy one.


ceciliasaraiva (approved):
nice, so happy to get rid of all these legacy code 🙌

emin-grbo (approved):

PR #175, opened 2024-11-29
[Player] Unified caching (2/X): Unified `CacheStorage` setup

asendra:
## Context

With the legacy Player implementation out of the picture, we can start migrating all `GenericMediaPlayer` implementations to use the new unified caching layer.

This will allow us to enforce a LIFO with maximum storage size quotas between all of them, and the possibility to inform of the storage usage and the ability to clear the cache as requested.

## Description

This PR only takes care of initializing the `CacheStorage`, and passing it to the relevant layers in preparation.


Tests/PlayerTests/Player/PlayerTests.swift
```diff
@@ -8,7 +8,8 @@ import XCTest
 final class PlayerTests: XCTestCase {
 	private var player: Player!
 	private var playerEventSender: PlayerEventSenderMock!
-	private var dbQueue: DatabaseQueue!
+	private var offlineDbQueue: DatabaseQueue!
+	private var cachedDQueue: DatabaseQueue!
```
ceciliasaraiva, line 12: [suggestion] small typo I guess to make it consistent with the above
```suggestion
	private var cachedDbQueue: DatabaseQueue!
```

ceciliasaraiva (approved):
nice, just one tiny suggestion

emin-grbo (approved):

PR #177, opened 2024-12-04
[Player] Unified caching (3/X): Better URL handling in GRDB

asendra:
## Context

While updating Player to use the new unified cache layer, I noticed a couple of issues with the current approach about storing and reading URLs from GRDB entities.

It is important to understand that URL are relative to the app sandbox fodler, and this can change after an OS update, a phone backup restore, etc

Due to this, we can't rely on the raw URL stored, and we need to use the `bookmarkData()` method available for this. 

### Update Cache entity

First, I changed the GRDB `CacheEntryEntity` to store the bookmark data instead of the URL, just like we do with the GRDB `OfflineEntryEntity`. We don't need a proper migration because the code has never run in production yet.

### Update bookmark data if stale

We were not detecting when the bookmark data was stale and updating the stored entity accordingly. This could cause issues in the future, so I made the necessary changes to fix it.

### DIFF
--- a/Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/OfflineEntryGRDBEntity.swift
+++ b/Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/OfflineEntryGRDBEntity.swift
@@ -56,8 +56,7 @@ struct OfflineEntryGRDBEntity: Codable, FetchableRecord, PersistableRecord {
 		var isStale = false
 		guard
 			let mediaBookmark,
-			let url = try? URL(resolvingBookmarkData: mediaBookmark, bookmarkDataIsStale: &isStale),
-			!isStale
+			let url = try? URL(resolvingBookmarkData: mediaBookmark, bookmarkDataIsStale: &isStale)
 		else {
 			return nil
 		}
@@ -68,12 +67,10 @@ struct OfflineEntryGRDBEntity: Codable, FetchableRecord, PersistableRecord {
 		var isStale = false
 		guard
 			let licenseBookmark,
-			let url = try? URL(resolvingBookmarkData: licenseBookmark, bookmarkDataIsStale: &isStale),
-			!isStale
+			let url = try? URL(resolvingBookmarkData: licenseBookmark, bookmarkDataIsStale: &isStale)
 		else {
 			return nil
 		}
-
 		return url
 	}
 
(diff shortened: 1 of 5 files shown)


Sources/Player/OfflineEngine/Internal/Storage/GRDBStorage/OfflineEntryGRDBEntity.swift
```diff
@@ -109,4 +106,22 @@ struct OfflineEntryGRDBEntity: Codable, FetchableRecord, PersistableRecord {
...
+			if let mediaBookmark {
+				_ = try URL(resolvingBookmarkData: mediaBookmark, bookmarkDataIsStale: &mediaBookmarkIsStale)
+			}
+			if let licenseBookmark {
+				_ = try URL(resolvingBookmarkData: licenseBookmark, bookmarkDataIsStale: &licenseBookmarkIsStale)
+			}
+
+			return mediaBookmarkIsStale || licenseBookmarkIsStale
+		} catch {
+			return false
```
ceciliasaraiva, line 124: [question] just checking in case it would be relevant to have logging here. Since we would get false in such case, and the continue flow would not succeed I suppose.

Sources/Player/PlaybackEngine/Internal/Cache/GRDBCache/CacheEntryGRDBEntity.swift
```diff
@@ -8,39 +8,64 @@ extension CacheEntryType: DatabaseValueConvertible {}
...
+		CacheEntry(key: key, type: type, url: mediaURL, lastAccessedAt: lastAccessedAt, size: size)
+	}
+
+	private var mediaURL: URL? {
+		var isStale = false
+		guard
+			let mediaBookmark,
+			let url = try? URL(resolvingBookmarkData: mediaBookmark, bookmarkDataIsStale: &isStale)
+		else {
+			return nil
```
ceciliasaraiva, line 27: [question] would it make sense to have some logging here?

Sources/Player/PlaybackEngine/Internal/Cache/GRDBCache/GRDBCacheStorage.swift
```diff
@@ -11,16 +11,24 @@ final class GRDBCacheStorage {
...
+					t.column("mediaBookmark", .blob)
 					t.column("lastAccessedAt", .datetime).notNull()
 					t.column("size", .integer).notNull()
 				}
 			}
+
+			try migrator.migrate(dbQueue)
+		} catch {
+			// TODO: Log error
+			print("Failed to initialize Cache database: \(error)")
```
ceciliasaraiva, line 30: [question] I guess here you do want logging 😅, not sure if you will add in this PR or future one.

ceciliasaraiva (approved):
only some minor comments

PR #182, opened 2025-01-09
Remove isUserLoggedIn check in TidalApi

ziad-halabi9:
Currently the `RequestHelper` checks for a few conditions before making requests. One of the check is whether the user is logged in. It throws an error if that's not the case. This limits the TidalAPI to be only used when the user is authenticated. 

This prevents its usage with Client Credentials auth flow where the client is authenticated and capable to make API requests that don't require user authentication. This PR removes the check of `isUserLoggedIn == true`.

### DIFF
--- a/Sources/TidalAPI/Utils/RequestHelper.swift
+++ b/Sources/TidalAPI/Utils/RequestHelper.swift
@@ -15,10 +15,7 @@ enum RequestHelper {
 		let credentials = try await credentialsProvider.getCredentials()
 		let requestBuilder = try await requestBuilder()
 		let requestURL = requestBuilder.URLString
-		guard
-			let token = credentials.token,
-			OpenAPIClientAPI.credentialsProvider?.isUserLoggedIn == true
-		else {
+		guard let token = credentials.token else {
 			throw TidalAPIError(
 				message: "NO_TOKEN",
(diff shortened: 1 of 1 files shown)


deaa (approved):

Sources/TidalAPI/Utils/RequestHelper.swift
```diff
@@ -15,10 +15,7 @@ enum RequestHelper {
 		let credentials = try await credentialsProvider.getCredentials()
 		let requestBuilder = try await requestBuilder()
```
asendra, line 16: Ultra-minor thing.

```let requestBuilder = try await requestBuilder()```

This line could be after the token check. No need for it before, so it would be a bit more optimal.

asendra (approved):
Minor comment, but the change I think it is fine

Sources/TidalAPI/Utils/RequestHelper.swift
ziad-halabi9: Yeah good catch, will follow up in another PR for this 👍🏻

ziad-halabi9: Actually it is before the check because `requestURL` is used in the error thrown.

PR #214, opened 2025-06-11
OpenAPI generator improvements

asendra:
* Fix the `modelOneOf` custom template to properly support discriminators in the schema
* Improved the generator script with a new cli option `--skip-download` so we can regenerate the code from a local schema (without redownloading the schema from the server)


asendra:
Backend is doing some more work on this, so we can't merge this until that is finished.

To unblock ourselves, I will be [merging another PR](https://github.com/tidal-music/tidal-sdk-ios/pull/215) with a manual patch in the generated code.

PR #253, opened 2025-09-29
iOS: Auto-resume after short Bluetooth route loss

asendra:
## Summary
- Auto-resume playback after short Bluetooth route loss
- Pause on route loss, remember if playback was active on BT
- Resume if BT returns within a short window (default 5s)
- Scope: AudioSessionRouteChangeMonitor

## Implementation
- Track last route loss time using TimeProvider timestamp (NTP-backed)
- Internal constant for window duration defined in the monitor for easy tuning; no public API change
- Introduced AudioInfoProvider.isBluetoothOutputRoute() and used it in the monitor

## Tests
- Added unit tests to simulate AVAudioSession.routeChangeNotification and exercise auto-resume logic:
  - Tests/PlayerTests/Player/PlaybackEngine/Internal/AudioSessionRouteChangeMonitorTests.swift
  - Mocks TimeProvider and AudioInfoProvider to control timing and route type
- These tests are compiled for iOS only (#if !os(macOS)).

## Notes
- Does not change behavior for long disconnects or route changes to non-Bluetooth outputs
- No public Configuration changes; all internal to the monitor

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,9 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
 ## [Unreleased]
-- Fix Player: stabilize AudioSessionRouteChangeMonitor tests and add macOS guard to extension.
+
+### Changed
+- Auto-resume playback after short conenction drops over Bluetooth. (Player)
 
 ## [0.7.2] - 2025-09-24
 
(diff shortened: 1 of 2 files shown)


bkdl:
Looks like a good improvement to me, needs to be tested before integrating. Tests seem to not compile.

asendra:
> @asendra Sorry for the confusion, talking about testing before integrating I meant manual playback testing. 🙂

Yeah I know! But tests were not compiling on this PR, which is what I meant

A bit tricky to test this. AirPods are not easy to disconnect/connect.

PR #270, opened 2025-10-20
Player: Narrow UC product type to local file playback and rename it LOCAL

asendra:
## Summary

Rebased onto current `main` (the branch was 684 commits behind and conflicting) and extended with the rename we discussed.

`UC` meant two unrelated things: *play these bytes off disk*, and *fetch this remote URL as HLS*. Only the first has ever been used — the sole construction site in the app passes a file URL, and `productUrl` is `nil` everywhere else. This drops the remote half and names what is left.

## Changes

- `ProductType.UC(url:)` → `ProductType.LOCAL(url:)`, documented as file-URL only. `rawValue` moves from `"UC"` to `"LOCAL"`.
- `UCMediaPlayer`, both `loadUC` implementations and the `loadUCFailed` log case are gone. A local file now loads through the same `AVURLAsset` path as a video.
- `credentialsProvider` drops out of `PlayerLoader` and `InternalPlayerLoader`. Its only user was the bearer-token fetch that ran before playing a file already on disk.
- `getUCPlaybackInfo` → `getLocalPlaybackInfo`, no longer `async` (it never awaited anything).

## Notes for review

- **This is a source-breaking change** for anyone pattern-matching `.UC`. The only known consumer is the TIDAL iOS app, which constructs `.UC(url:)` in exactly one place (`BoomBoxTypeConverter.productType(from:)`).
- **`rawValue` is deliberately changed**, not pinned. It reaches the backend via `PlayLogEvent`, `PlaybackStatistics`, `DownloadStatistics` and `PlayerItem.sessionProductType`. There is no existing `"UC"` traffic to preserve — the path has never fired in production — so nothing regresses, but flag it if analytics wants `"UC"` kept.
- **Behaviour improves for file URLs.** The deleted `loadUC` attached `AVURLAssetHTTPHeaderFieldsKey` (useless for a file URL) and never set `AVURLAssetPreferPreciseDurationAndTimingKey`. The generic `load(_ url:cacheKey:…)` sets it based on `url.isFileURL`.
- **The fabricated playback info is unchanged** — `audioQuality: .LOW`, `audioCodec: .HE_AAC_V1`, `mediaType: HLS` are still placeholders rather than measurements of the file. Left alone to keep this diff to the rename; worth a follow-up.
- `Offliner` already returns `nil` for this case, so offlining is unaffected.

## Attribution

The narrowing is @asendra's original work from this PR (Oct 2025), re-applied onto `main` rather than merged — resolving eight conflicts across a 684-commit gap risked silently reverting a year of changes in those files. Their original five commits are preserved at tag `backup/cleanup-uc-type-20251022`.

## Verification

`swift build` clean for `Player` and `Offliner`, `--build-tests` compiles, swiftformat and swiftlint clean on the changed files.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+- **Breaking:** renamed `ProductType.UC(url:)` to `ProductType.LOCAL(url:)`, and its `rawValue` from `"UC"` to `"LOCAL"`. The case is now documented as file-URL only. (Player)
+
+### Removed
+- Removed the remote half of the former UC path: the `UCMediaPlayer` protocol, both `loadUC` implementations and the `loadUCFailed` log case. A local file now loads through the same `AVURLAsset` path as a video, which sets `AVURLAssetPreferPreciseDurationAndTimingKey` for file URLs. (Player)
+- Removed `credentialsProvider` from `PlayerLoader` and `InternalPlayerLoader`; the deleted auth-token fetch was its only user. (Player)
+
 ## [0.12.6] - 2026-08-24
 
 ### Changed
--- a/Sources/Offliner/Offliner.swift
+++ b/Sources/Offliner/Offliner.swift
@@ -394,7 +394,7 @@ extension Offliner: OfflineItemProvider {
 		switch productType {
 		case .TRACK: mediaType = .tracks
 		case .VIDEO: mediaType = .videos
-		case .UC: return nil
+		case .LOCAL: return nil
 		}
 
 		do {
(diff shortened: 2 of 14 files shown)


bkdl (approved):
Nice

deaa:
Don't we still using this as playing local files?

asendra:
Ahh… you mean for playing the uploaded files locally? I guess yeah, that is
true? 🤔 🥲


For offline files with the legacy system we use a different path with
StoredMediaProduct.

asendra:
@bkdl @deaa I've restored the minimum necessary code needed for local playback of those files.

bkdl:
🤖 (sent by Even's AI agent)

Closing this in favour of #450.

This branch had been force-pushed to a rebased-and-extended version of the change, which left the approval here attached to a diff that no longer existed — and `main` does not dismiss stale reviews. Rather than merge under that, #450 carries the same change on a fresh branch off current `main` for a fresh review.

@deaa your catch about local file playback is carried over and called out explicitly in the new description — that is the reason the change narrows-and-renames rather than deletes.

@asendra the narrowing is your work and is credited as such. Your original commits from this branch are preserved at tag `backup/cleanup-uc-type-20251022`.

PR #369, opened 2026-04-17
Player: Add fade-out support to pause for sleep timer

deaa:
## Summary
- Extends `pause()` with an optional `fadeDuration: TimeInterval?` parameter across the full player stack: `GenericMediaPlayer` → `AVQueuePlayerWrapper` → `CrossfadingPlayerWrapper` → `PlayerItem` → `PlayerEngine` → `Player`
- When a fade duration is provided, applies an `AVMutableAudioMix` volume ramp from 1.0 to 0 over the specified duration, then pauses and restores volume — reuses the same `setVolumeRamp` mechanism as crossfade
- A protocol extension provides the default `pause()` → `pause(fadeDuration: nil)` for backward compatibility

## Context
This enables the sleep timer feature in the TIDAL iOS app to fade out audio smoothly before pausing, instead of an abrupt stop.

## Test plan
- [ ] Verify normal `pause()` behavior is unchanged (no fade)
- [ ] Verify `pause(fadeDuration: 8)` produces a smooth 8-second fade-out then pauses
- [ ] Verify volume is restored to its original level after pause so next play starts normally
- [ ] Verify crossfade between tracks still works correctly
- [ ] Verify fade works with both FLAC and AAC streams

### DIFF
--- a/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
+++ b/Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
@@ -167,18 +167,8 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
 				return
 			}
 
-			let currentTime = self.player.currentTime()
-			let timeRange = CMTimeRange(
-				start: currentTime,
-				duration: CMTime(seconds: fadeDuration, preferredTimescale: 600)
-			)
-
-			let inputParameters = AVMutableAudioMixInputParameters()
-			inputParameters.setVolumeRamp(
-				fromStartVolume: 1.0,
-				toEndVolume: 0,
-				timeRange: timeRange
-			)
+			let timeRange = self.getFadeTimeRange(fadeDuration: fadeDuration)
+			let inputParameters = self.getVolumeRampInputParameters(timeRange: timeRange)
 
 			let audioMix = AVMutableAudioMix()
 			audioMix.inputParameters = [inputParameters]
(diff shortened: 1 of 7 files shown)


Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
```diff
@@ -160,9 +160,38 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
...
-	func pause() {
+	func pause(fadeDuration: TimeInterval? = nil) {
 		queue.dispatch {
-			self.player.pause()
+			guard let fadeDuration, fadeDuration > 0, let playerItem = self.player.currentItem else {
+				self.player.pause()
+				return
+			}
+
+			let currentTime = self.player.currentTime()
```
bkdl: Could be extracted to a `getFadeTimeRange` function

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
```diff
@@ -160,9 +160,38 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
...
+			let timeRange = CMTimeRange(
+				start: currentTime,
+				duration: CMTime(seconds: fadeDuration, preferredTimescale: 600)
+			)
+
+			let inputParameters = AVMutableAudioMixInputParameters()
```
bkdl: Could be extracted to a `getVolumeRampInputParameters(timeRange:) function`

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
```diff
@@ -160,9 +160,38 @@ final class AVQueuePlayerWrapper: GenericMediaPlayer {
 		}
 	}
 
-	func pause() {
+	func pause(fadeDuration: TimeInterval? = nil) {
```
bkdl, line 163: We could also introduce an `Effect` enum with `case .fadeOut(duration:)`, but maybe a bit overkill

deaa: Agree, overkill for now — it's a single use case. If we ever need `.fadeIn` we can introduce it then.

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
deaa: Done in dea3d0f — extracted `getFadeTimeRange` and `getVolumeRampInputParameters`.

Sources/Player/PlaybackEngine/Internal/AVQueuePlayer/AVQueuePlayerWrapper.swift
deaa: Done in the same commit 👆

bkdl (approved):

PR #399, opened 2026-06-11
TM-1291: Stream collection download state

bkdl:
## Why
The iOS client needs album, playlist, and mix download buttons to reflect Cloud Offliner media availability without stale one-shot snapshots or misleading removal/download states.

## What
- Add collection download-state streaming semantics backed by local store, active downloads, and offliner tasks
- Emit a fast local first value, then reconcile against backend task activity
- Treat SDK-initiated removals as locally not downloaded immediately
- Keep removal tasks out of the `.downloading` state

## Risk Assessment
Medium-low — scoped to the Offliner collection download-state API and related tests/docs. The first emission is intentionally local-first and may be reconciled by backend task state shortly after.

## References
- Tests not run locally; pre-commit SwiftLint environment previously failed during dependency resolution before linting repo code.

Generated with Codex


Sources/Offliner/Internal/OfflineApiClient.swift
```diff
@@ -107,6 +140,31 @@ protocol OfflineApiClientProtocol {
...
 extension OfflineApiClientProtocol {
+	func getCollectionTaskActivity(
+		collectionType: OfflineCollectionType,
+		resourceId: ResourceId
+	) async throws -> OfflineCollectionTaskActivity {
+		let collection = OfflineCollectionReference(collectionType: collectionType, resourceId: resourceId)
+		var activity = OfflineCollectionTaskActivity.none
+		var cursor: String?
+
+		repeat {
```
tidal-gaute: This concerns me in terms of backend load. It's hard for backend to protect itself since this isn't cacheable (the option is rate limiting, and I'm uncomfortable with designing something that probably requires rate limiting). For multiple clients with large task-lists, this will be a very heavy on backend.

If we really want to say "not downloaded" when it's in the process of being deleted. Then I think we should add some metadata to the response of getOfflineCollection where backend states what "activity" is ongoing for that collection.

If we remove this from now, we're still getting the benefits of the button working "properly" when you download it, right? Or has it always done that?

bkdl: 🤖 Updated this in `345a466a` by removing the SDK-side `getCollectionTaskActivity` / `/offlineTasks` scan entirely, so the stream no longer adds backend load on view entry.

The download button still works for work known by this SDK instance: pending/running collection metadata tasks and member media downloads are treated as `.downloading`; removal optimistically deletes the local collection row so state becomes `.notDownloaded`.

The remaining limitation is cold-start correctness for backend tasks created outside this SDK instance. For that, I agree the right backend shape would be per-collection activity metadata on `getOfflineCollection` rather than client-side task-list scanning.

tidal-gaute (approved):

PR #425, opened 2026-08-01
Fix Offliner compile errors after API regeneration

deaa:
## Context

#419 (the `tidal-music-tools/auto-update-tidal-api` regeneration) is failing `Run Swift Unit Tests`: the regenerated models broke the hand-written `Offliner` module, so the SPM build dies at compile time before any test runs. This PR targets that branch and fixes only the two hand-written files — no generated content is touched.

## Changes

The regeneration made `ArtworkFile.meta` non-optional (with non-optional `width`/`height`), split `ResourceIdentifier` into `OfflineTasksItemResourceIdentifier` / `OfflineTasksCollectionResourceIdentifier`, and specialized the albums relationship into `TracksAlbumsMultiRelationshipDataDocument` / `VideosAlbumsMultiRelationshipDataDocument`.

- `ArtworkDownloader.swift`: drop optional chaining in the `area` computation
- `OfflineApiClient.swift`: adopt the specialized identifier types in task payload helpers; `wireRelationships` now takes the album as a `(type, id)` tuple since tracks and videos use distinct albums document types

Verified locally with `swift build --target Offliner`.

### DIFF
--- a/Sources/Offliner/Internal/ArtworkDownloader.swift
+++ b/Sources/Offliner/Internal/ArtworkDownloader.swift
@@ -45,6 +45,6 @@ enum ArtworkDownloaderError: Error {
 
 private extension ArtworkFile {
 	var area: Int {
-		(meta?.width ?? 0) * (meta?.height ?? 0)
+		meta.width * meta.height
 	}
 }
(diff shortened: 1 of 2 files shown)


bkdl (approved):
Looks good to me

PR #432, opened 2026-08-06
Offliner: Recover orphaned download tasks and simplify task fetching

tidal-gaute:
## Summary

**Orphaned download task recovery.** When the background `AVAssetDownloadURLSession` reports a task the `MediaDownloader` no longer tracks (e.g. the app was killed mid-download), the task previously just leaked. `MediaDownloader` now exposes an `orphanedTaskHandler` closure, invoked from `didCompleteWithError` (`didFinishDownloadingTo` is not reliably called for reconnected tasks). `Offliner` wires it up to mark the task `FAILED` on the backend and rerun the `TaskRunner` so the download is retried. The update payload only supports `IN_PROGRESS`/`FAILED`/`COMPLETED`, so `FAILED` is used to re-queue.

**Cursorless task fetching.** `TaskRunner` no longer paginates with a cursor. It fetches the oldest page of pending tasks and only refreshes when fewer than 10 tasks are pending locally. Also fixes a stall where tasks fetched on the last completion in the processing loop were never started: the loop now refreshes before picking the next task.

## Testing

- All OfflinerTests pass (74/74)
- Verified manually in the iOS app against a local SDK checkout: orphaned tasks are re-queued and downloads resume after app relaunch


Sources/Offliner/Internal/TaskRunner.swift
```diff
@@ -154,12 +149,12 @@ actor TaskRunner {
 			}
 
 			for await _ in group {
-				if let task = getTask() {
-					group.addTask { await self.start(task) }
+				if pendingTasks.count < Self.refreshThreshold {
+					try? await refresh()
```
bkdl, line 153: Just noticed that we're just swallowing errors here. Not sure there's anything worthwhile to do to handle them now, but something to keep in mind.

bkdl (approved):

Sources/Offliner/Internal/TaskRunner.swift
tidal-gaute: We could print something, but it will retry on the next iteration.

PR #433, opened 2026-08-08
TidalAPI: add body-ignoring create dry-run APIs

stefanmoro:
## Summary

- Add explicit `artistsPostDryRun` and `appreciationsPostDryRun` APIs that force `meta.dryRun = true`.
- Reuse the generated request builders, authentication, and retry behavior while intentionally ignoring successful response bodies.
- Cover minimal JSON:API success documents returned with either `200` or `201`, while continuing to reject non-2xx responses.

## Scope

This is the minimal client-side compatibility change needed before the server can return a generic `{ links, meta }` document for create dry runs. It does not edit the SDK-local OAS, regenerate endpoint signatures, or document/add `204`. Ordinary non-dry-run create methods remain typed and strict.

## Verification

- `swift build --target TidalAPI`
- `swiftformat --lint` on changed Swift files
- Focused unit tests run in CI

### DIFF
--- a/Package.swift
+++ b/Package.swift
@@ -77,6 +77,12 @@ let package = Package(
 				"README.md",
 			]
 		),
+		.testTarget(
+			name: "TidalAPITests",
+			dependencies: [
+				.tidalAPI,
+			]
+		),
 		.target(
 			name: "Common",
 			dependencies: [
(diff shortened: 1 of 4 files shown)


stefanmoro:
🤖 Closing this approach because it adds a transitional 204 response to the SDK-local OAS to generate optional return types. The authoritative server OAS does not contain that response, and routine regeneration would remove both the OAS edit and the generated signatures. There is no minimal code-only way for the existing non-optional typed create methods to accept a generic success document. We should revisit iOS compatibility when the final 200/201 response contract is defined in the authoritative OAS.

stefanmoro:
🤖 Closing in favor of shipping the server contract first, then regenerating the SDK from the published OAS so any compatibility change is driven by the real 200/201 response shapes.

PR #437, opened 2026-08-18
TM-1767: Stabilize collection download state

bkdl:
### Track download intent inside the SDK

Collection downloads are registered with TaskRunner before the backend submission begins. This closes the gap where a newly opened observer could report notDownloaded after the user had already started a download but before the backend task appeared in TaskRunner.

TaskRunner distinguishes two request phases:

1. requested — the command has started but no related SDK task has been observed yet
2. acknowledged — TaskRunner has observed related acquisition work

The request remains active until its related tasks finish. Submission failure and successful removal clear it explicitly.

### Make local availability stable

Collection state is now resolved in this order:

1. A user-requested download that has not finished reports downloading.
2. A locally stored collection reports downloaded.
3. Other active related acquisition work reports downloading only when the collection is not locally stored.
4. Otherwise the collection reports notDownloaded.

This prevents an already-offline collection from temporarily changing to downloading during background reconciliation, without reporting a newly requested download as complete merely because collection metadata was stored before its remaining item tasks finished.

### Regression coverage

- request state is visible while backend submission is still suspended
- locally available content stays downloaded during unrelated related work
- failed submission clears the pending request
- existing state-stream behavior remains covered

## Consumer impact

This lets the iOS app use the SDK stream as the durable source of truth across navigation. The app only needs a transient optimistic button transition for immediate tap feedback; it no longer needs its own download-state coordinator or a second local-availability query.

## Mechanical formatting

The repository pre-commit SwiftFormat hook reformatted the touched files. Changes such as removing redundant self references, normalizing enum case patterns, wrapping guards, and adding MARK separators are formatter-only and do not affect behavior.

## Validation

- pre-commit SwiftFormat passed
- pre-commit SwiftLint passed
- tests not run locally

### DIFF
--- a/Sources/Offliner/Internal/TaskRunner.swift
+++ b/Sources/Offliner/Internal/TaskRunner.swift
@@ -2,7 +2,14 @@ import Foundation
 import Network
 import OSLog
 
+// MARK: - TaskRunner
+
 actor TaskRunner {
+	private enum CollectionDownloadRequestState {
+		case requested
+		case acknowledged
+	}
+
 	private static let logger = Logger(subsystem: "com.tidal.sdk.offliner", category: "TaskRunner")
 	private static let maxConcurrentTasks = 5
 	private static let refreshThreshold = 10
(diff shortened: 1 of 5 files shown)


bkdl:
🤖 Superseded by #445, which resolves collection download state from the backend inventory instead of relying on local transient state (that approach couldn't stay correct across app restarts or for collections queued behind a large download). The app side (tidal-engineering/iOS#8832) now pins this SDK's `agent/offliner-backend-driven-download-state` branch, and the playlist download flow was verified end-to-end against it. Closing in favor of #445.

PR #445, opened 2026-08-19
Offliner: Resolve collection download state from the backend inventory

tidal-gaute:
## What

Replaces the polling-based `getOfflineCollectionDownloadState` with backend-driven resolution:

- The stream emits the local storage state first (works offline, instant), then a backend-corrected state when the offline inventory is reachable.
- Resolution order: backend reports pending work → `.downloading`; local collection row exists → `.downloaded`; otherwise `.notDownloaded`. If the backend is unreachable, the last emitted state stands.
- `TaskRunner` is trigger-only: when a task related to an observed collection starts or finishes, or the collection is downloaded/removed, it signals the stream to re-resolve against the backend. Its in-memory task list is never used as state truth.

## Why

The in-memory approach can't be correct: the backend queue interleaves tasks from many collections and is paginated, so locally invisible pending work exists behind the first page, and in-memory state doesn't survive restarts. The backend already resolves a collection's pending/stored state transactionally, so asking it directly removes the race conditions entirely instead of papering over them.

This supersedes the approach in #437 (TM-1767).

## How

- `TaskRunner.collectionEvents(for:)` returns an `AsyncStream<Void>` per observed collection; `refresh()`, `finish()`, and `notifyCollectionChanged` yield to matching observers. `bufferingNewest(1)` coalesces event bursts into a single refetch.
- Removed the poll interval, the polling loop, and `hasCurrentDownload`.

### DIFF
--- a/Sources/Offliner/Internal/TaskRunner.swift
+++ b/Sources/Offliner/Internal/TaskRunner.swift
@@ -21,6 +21,7 @@ actor TaskRunner {
 	private var pendingTasks: [InternalTask] = []
 	private var runningTasks: [InternalTask] = []
 	private var taskIds: Set<String> = []
+	private var collectionObservers: [UUID: (collection: OfflineCollectionReference, continuation: AsyncStream<Void>.Continuation)] = [:]
 
 	private var processTask: Task<Void, Never>?
 	private var rerunRequested = false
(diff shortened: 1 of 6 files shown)


bkdl (approved):
Seems to work great. 👍

